PR #27, opened 2023-04-27
add checks for X

rcannood:


R/InMemoryAnnData.R
```diff
@@ -17,14 +17,15 @@ InMemoryAnnData <- R6::R6Class("InMemoryAnnData",
       if (missing(value)) {
         private$.X
       } else {
-        private$.X <- value
+        private$.X <- .inmemoryanndata_check_matrix(value, self$obs, self$var)
```
mtmorgan, line 20: It seems https://adv-r.hadley.nz/r6.html like the convention is to have a `validate()` member function; maybe this should be a member function `.validate_matrix()` ?

rcannood: I see, good point!

I'll tackle these changes in the `names_in_slot` branch, since that branch superseded this one.

rcannood:
Superseded by #29

PR #28, opened 2023-04-27
add seurat converter

rcannood:
Adds a `to_Seurat(...)` function


R/to_seurat.R
```diff
@@ -0,0 +1,17 @@
+#' Convert an AnnData object to Seurat
+#' 
+#' @param obj An AnnData object
+#' 
+#' @importFrom methods is
+#' @importFrom Seurat CreateAssayObject AddMetaData CreateSeuratObject
```
mtmorgan, line 6: I guess there should be a `#' @export`. Also I find having a light-weight `#' @examples` (e.g., create a small InMemoryAnnData and the coerce to Seurat). It's a light-weight unit test that's visible to the user.

NAMESPACE
```diff
@@ -2,3 +2,7 @@
 
 importFrom(Matrix,as.matrix)
 importFrom(R6,R6Class)
+importFrom(Seurat,AddMetaData)
+importFrom(Seurat,CreateAssayObject)
+importFrom(Seurat,CreateSeuratObject)
```
mtmorgan, line 7: you don't want to import things from Seurat, unless Seurat is in the Imports: field. And you don't want Seurat in the Imports: field because the users of SingleCellExperiment then have a very heavy set of dependencies, and vice-versa. I think the code you've written, with `Seurat::...` does not require these import statetements.

R/to_seurat.R
```diff
@@ -0,0 +1,17 @@
...
+to_seurat <- function(obj) {
```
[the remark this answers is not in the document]
rcannood: Fair enough. Done!

R/to_seurat.R
rcannood: Good idea, done!

NAMESPACE
rcannood: Good point, my bad.

rcannood: Fixed :)

PR #29, opened 2023-04-27
Move obs and var names to separate slot

rcannood:
Since the obs_names and the var_names might soon be allowed to be non-character objects, we're storing them as a separate field in the in memory data structure

### DIFF
--- a/tests/testthat/test-inmemory-anndata.R
+++ b/tests/testthat/test-inmemory-anndata.R
@@ -1,26 +1,26 @@
 library(Matrix)
 
+# construct dummy X
+X <- Matrix::rsparsematrix(nrow = 10, ncol = 20, density = .1)
+
+# construct dummy obs
+obs <- data.frame(
+  cell_type = sample(c("tcell", "bcell"), 10, replace = TRUE),
+  cluster = sample.int(3, 10, replace = TRUE)
+)
+
+# construct dummy obs names
+obs_names <- paste0("cell_", seq_len(10))
+
+# construct dummy var
+var <- data.frame(
+  geneinfo = sample(c("a", "b", "c"), 20, replace = TRUE)
+)
+
+# construct dummy var names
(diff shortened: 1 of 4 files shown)


tests/testthat/test-inmemory-anndata.R
```diff
@@ -35,22 +41,61 @@ test_that("InMemoryAnnData$new() fails gracefully", {
...
+      var_names = var_names
+    )
+  })
 
-test_that("dim() works", {
-  ad <- InMemoryAnnData$new(shape = c(1L, 1L))
-  expect_identical(ad$shape(), c(1L, 1L))
-
-  ## from obs and var, or if that fails X
-  ad <- InMemoryAnnData$new(obs = data.frame(x = 1:3), shape = c(3L, 0L))
```
mtmorgan, line 50: I wonder if it makes sense for `new()` to 'do the right thing' if it has sufficient memory, like here replacing the NULL X with `matrix(0, 3, 0)` and var with `data.frame()`, so no requirement for `shape`? This would also make `new()` with no arguments create a valid object.

rcannood: In Python, it's possible to create AnnData objects where X is None (i.e. NULL).

```python
import anndata as ad
adata = ad.AnnData(shape = [3,1])
adata.X is None # True
```

This is because the AnnData object might have layers with dimensions instead, or might only contain obs, var, obsm and varm metadata and no expression matrix at all.

So I wouldn't replace `X` with `matrix(0, nrow = nrow(obs), ncol = nrow(var))` or even an empty sparse matrix, since it would deviate from the anndata implementation.

mtmorgan: OK, so `InMemoryAnnData$new(shape = c(3, 5))` should / does work, including for the special case `InMemoryAnnData$new(shape = c(0, 0))`. Sounds good!

PR #36, opened 2023-04-28
Add _keys functions

mtmorgan:


R/AbstractAnnData.R
```diff
@@ -68,6 +68,14 @@ AbstractAnnData <- R6::R6Class("AbstractAnnData",
...
       nrow(self$var)
     },
+    #' @description Keys ('column names') of `obs`.
+    obs_keys = function() {
+      .abstract_function()
+    },
+    #' @description Keys ('column names') of `var`.
+    var_keys = function() {
+      .abstract_function()
+    },
```
rcannood, line 78: Can we not implement this in the abstract class?

```suggestion
    obs_keys = function() {
      names(self$obs)
    },
    #' @description Keys ('column names') of `var`.
    var_keys = function() {
      names(self$var)
    },
```

R/InMemoryAnnData.R
```diff
@@ -162,6 +162,14 @@ InMemoryAnnData <- R6::R6Class("InMemoryAnnData",
...
       private$.X <- private$.validate_matrix(X, "X")
       private$.layers <- private$.validate_layers(layers)
+    },
+    #' @description `obs_keys()` returns the keys ('column names') of `obs`.
+    obs_keys = function() {
+      names(private$.obs)
+    },
+    #' @description `var_keys()` returns the keys ('column names') of `var`.
+    var_keys = function() {
+      names(private$.var)
```
rcannood, line 172: If the other suggestion is accepted.

```suggestion
```

rcannood (changes requested):

R/AbstractAnnData.R
mtmorgan: Is it guaranteed (or 'expected') that self$var returns something with `names()`?

rcannood: Indeed :)

rcannood (approved):

PR #40, opened 2023-04-29
refactor converters

rcannood:


R/AbstractAnnData.R
mtmorgan, line 1: It seems like this functionality isn't the responsibility of the abstract class, but rather of the derived classes -- `to_InMemoryAnnData()` as a plain-old-function in R/InMemoryAnnData.R?

R/to_SingleCellExperiment.R
mtmorgan, line 1: What if there were a `from_SingleCellExperiment()`? I would rather the modularity of all SCE functionality in one file.

I like the idea that every R/ file has a corresponding man/ and tests/testhat/ file. So renaming this would also rename the documentation module and unit test file.

rcannood: I see. Makes sense that everything would be split in the same way. I'll do it the other way around then; I'll rename to_Seurat to simply Seurat.R.

PR #41, opened 2023-04-29
Add lintr

rcannood:
* Add `lintr` to CI
* Ran `styler::style_pkg()` to fix some issues

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -6,3 +6,4 @@
 ^README\.qmd$
 ^\.metals$
 ^\.vscode$
+^\.lintr$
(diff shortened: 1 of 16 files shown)


rcannood:
Hmm. `lintr` expects variable names to be snake case or symbols:

> Variable and function name style should be snake_case or symbols.

Should we switch everything inside the package to snake_case? 

Example:

```r
adata <- AnnData( ... ) # <- exception?
sce <- adata$to_sce()
seu <- adata$to_seurat()
som <- adata$to_soma()
```

mtmorgan:
I feel (strongly) that classes should be UpperCamelCase.

mtmorgan:
I'll address many of these as part of https://github.com/scverse/anndataR/pull/42

rcannood:
> I feel (strongly) that classes should be UpperCamelCase.

Yeah, I agree about the classes.

Imo I agree with most of the linting, except:

* Classes should be CamelCase
* X as a variable should be allowed, given the context

I'm on the fence about `to_SingleCellExperiment` and `to_Seurat` as it mixes snake_case and CamelCase. However, `to_seurat and `to_sce` are also not very satisfying...

PR #50, opened 2023-05-03
Add remaining slots to AbstractAnnData

lazappi:
Add remaining slots to `AbstractAnnData` so that these features can be implemented in the other interfaces. There are a few things to consider before this is merged:

**Raw**

I haven't a slot for `.raw` yet because I think it needs some more thought, we might need another class for this and/or to modify the existing classes. Not sure if we should deal with that now or leave it for another PR.

**Interface inconsistencies**

Not sure how important this is but there are some inconsistencies in what we have implemented as "attributes" vs "methods" (if you consider `active` as equivalent to a Python attribute and `public` as equivalent to a Python method).

- `n_obs`, `n_vars`, `shape` are Python attributes but we have them in `public`.
- We have `layer_keys` which doesn't exist in Python (at least not in the documentation)

### DIFF
--- a/R/AbstractAnnData.R
+++ b/R/AbstractAnnData.R
@@ -27,26 +27,12 @@ AbstractAnnData <- R6::R6Class("AbstractAnnData",
     obs = function(value) {
       .abstract_function()
     },
-    #' @field var A `data.frame` with columns containing information
-    #'   about variables. The number of rows of `var` defines the variable
-    #'   dimension of the AnnData object.
-    var = function(value) {
-      .abstract_function()
-    },
     #' @field obs_names Either NULL or a vector of unique identifiers
     #'   used to identify each row of `obs` and to act as an index into
     #'   the observation dimension of the AnnData object. For
     #'   compatibility with *R* representations, `obs_names` should be a
     #'   character vector.
     obs_names = function(value) {
-      .abstract_function()
-    },
-    #' @field var_names Either NULL or a vector of unique identifiers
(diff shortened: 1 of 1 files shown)


mtmorgan:
I added `layer_keys()` so that I could implement both the `AbstractAnnData$print()` method and `to_SingleCellExpermient()` in a way that did not necessarily require full access to the object, and helped with abstraction.  For instance the `print` method has https://github.com/scverse/anndataR/blob/7234db468db08535083063d9c2ff91381d97e44f/R/AbstractAnnData.R#L68  which accesses the keys without having to, e.g., load all layers into memory as an R list and then do `names()`. A similar consideration allows `to_SingleCellExperiment()` to create `assays` by iterating through `layers[[<key>]]`, without having to know how to coerce `layers` to an R list (which would require each class to implement an `as.list()` method or something) https://github.com/scverse/anndataR/blob/7234db468db08535083063d9c2ff91381d97e44f/R/SingleCellExperiment.R#L34-L49 Writing it this way allows the same code to work with all (i.e., InMemoryAnnData and ReticulateAnnData) currently implemented classes.

rcannood:
I would be inclined to wait to merge this PR until all we've implemented all the 0.1.0 items. Would that be okay?

**Raw**: I agree that another class is needed for this.

**Inconsistenties**: if there are methods which do not exist in the python interface but do exist in ours, we could specifically mention this in the documentation or make them private. Either would be okay for me :)

Issue #51, opened 2023-05-03
Interface for writing to disk


lazappi:
We talked a bit about how the classes should be structured and how the interface to existing files should work but not so much about how writing a new file should work. We want users to be able to write a `SingleCellExperiment` (and maybe a `Seurat`???) to disk but there are different ways that could work. It would be nice to decide soonish before we get too far into implementing things. Possible options (I'm sure there are more I haven't thought of):

**Option 1:** Reverse conversion

Functions/methods that match the conversion to R objects but in reverse. Something like:

```r
interface <- to_AnnData(sce, interface = "InMemory"/"H5AD"/"Zarr"...)
```

Then the interface could have a writing method:

```r
interface$write("path/to/file")
```

This somewhat matches the Python interface but probably exposes the intermediate objects to normal users too much which I think we want to avoid.

***Option 1B***

You could also probably do something like this where the method acts on the abstract class (although I'm not sure exactly how that would work)

```r
interface <- to_AnnData(sce)
interface$write_h5ad("path/to/file")
```

**Option 2:** Direct reading

The other object is to have explicit `writeH5AD()`/`writeZarr()`... functions that act on `SingleCellExperiment`/`Seurat` objects (and maybe also the interfaces?). Something like:

```r
writeH5AD(sce, "path/to/file")
```

This could work directly on the object or maybe by hiding Option 1:

```
writeH5AD <- function(sce) {
	interface <- to_AnnData(sce, interface = "H5AD")
	interface$write("path/to/file")
}
```

We would probably want matching `readH5AD()` etc. functions as well. I think this interface is nicer/easier for users but maybe a bit more work because it doesn't match as well with the classes (particularly the in-memory ones).

**Option 3:** Some combination of these/something else

Thoughts?

rcannood:
I initially thought we would have a function called `from_SingleCellExperiment` and `from_Seurat`, e.g.

Originally, I'd planned to have these functions simply return an InMemoryAnnData:

```r
from_SingleCellExperiment <- function(sce) {
  # construct X, obs, var, ...
  InMemoryAnnData$new(...)
}

sce <- ...
adata <- from_SingleCellExperiment(sce)
h5ad <- adata$to_HDF5()
h5ad$write("dataset.h5ad")
```

However, it does make sense to have an argument to automatically perform this transformation for you:

```r
from_SingleCellExperiment <- function(sce, output_class = c("InMemory", "HDF5", "Reticulate")) {
  output_class <- match.arg(output_class)

[...]

PR #65, opened 2023-05-04
initial implementation for creating empty objects

rcannood:
Perhaps needed for #51


rcannood:
an alternative implementation was merged in #70

PR #67, opened 2023-05-04
HDF5 reading functions

LouiseDck:
Functions to read certain datatypes from the hdf5 files.

**Tests need to be added for the following reading functions:**
- [x] mapping
- [x] dataframe
- [ ] nullables
- [ ] categoricals

**Following bugs still need fixing:**

- [x] String arrays are not 2d matrices when they should be
- [x] A dataframe with no columns does need to have the right amount of rows


R/HDF5-read.R
```diff
@@ -159,8 +159,12 @@ read_h5ad_nullable_integer <- function(file, name, version = c("0.1.0")) {
 #' @return a character vector/matrix???
 read_h5ad_string_array <- function(file, name, version = c("0.2.0")) {
   version <- match.arg(version)
-  
-  rhdf5::h5read(file, name)
+  # reads in transposed
+  string_array <- rhdf5::h5read(file, name)
+  if(is.matrix(string_array)){
```
rcannood, line 164:
```suggestion
  if (is.matrix(string_array)) {
```

lazappi:
> * [ ]  A dataframe with no columns does need to have the right amount of rows

This I have already done in `hdf5-structure`. Possibly the cause of the conflict above.

lazappi:
@LouiseDck @rcannood What are your thoughts on merging this into `hdf5-structure` as is and opening smaller PRs for anything that still needs to be done? I'm finding all the branches hard to manage and most of this has already been merged there anyway.

rcannood:
I think these PRs are for you to merge ;) When Louise and I are working on something we're doing it in separate branches to ensure we don't cause any conflicts in your ongoing work.

PR #68, opened 2023-05-04
Hdf5-writing

rcannood:
* remove unused imports
* create factor only using levels and codes
* minor fixes for lintr
* worked on implementing HDF5 writing functions. Haven't been tested yet >_>


rcannood:
Merging this PR. It doesn't work yet, but the neither is the code in the `hdf5-structure` branch. However the round-trip works quite well.

PR #74, opened 2023-05-08
Update example dataset

lazappi:
The example Krumsiek H5AD files from **{zellkonverter}** we were using for testing were missing some things we want to check. This dataset should be more complete and easier to add things to as needed.


lazappi:
Possibly we also want a similar R example (maybe using the in-memory interface)? I wasn't sure how best to do that but if we come up with something it could go here or in a separate PR.

rcannood (changes requested):
Makes sense to remove the krumsiek dataset.

However, the resulting `example.h5ad` is too big for my liking. Can you create something that's a little smaller, e.g. <100KB?

lazappi:
Maybe??? It's only 100 cells x 200 genes so I'm not sure that can be reduced much. Some of the things might be redundant (in that they are the same types) but I'm not sure exactly which parts are taking up space. Any idea if there is an easy way to tell that from a HDF5 file?

I guess the other option would be to put it on Figshare and use **{BiocCache}** to get it as needed but it would be simpler to have something in the package.

lazappi:
Ok, I got it down to ~400kb. Small enough?

rcannood (approved):
400KB is already a lot better :)

lazappi:
FYI I want to add some things to this so please don't merge yet

R/HDF5-read.R
```diff
@@ -7,7 +7,6 @@
 #'
 #' @return A named list with names type and version
 read_h5ad_encoding <- function(file, name) {
-  requireNamespace("rhdf5")
```
rcannood, line 10: Isn't it better to keep `requireNamespace("rhdf5")` in here, if these functions are meant to be transferred to a separate package?

This applies to all of the other `read_h5ad_*` and `write_h5ad_*` functions as well.

R/HDF5-read.R
```diff
@@ -149,6 +148,29 @@ read_h5ad_sparse_array <- function(file, name, version = "0.1.0",
...
+#' Read H5AD recarray
+#'
+#' Read a recarray from an H5AD file
+#'
+#' @param file Path to a H5AD file or an open H5AD handle
+#' @param name Name of the element within the H5AD file
+#' @param version Encoding version of the element to read
+#'
+#' @details
+#' A "record array" (recarray) is a Python NumPy array type that countains
```
rcannood, line 160: Fix typo

```suggestion
#' A "record array" (recarray) is a Python NumPy array type that contains

R/HDF5-read.R
lazappi: I'm not quite sure what you mean? These functions aren't exported so I was imagining anyone depending on this would use `HDF5AnnData` which now contains a dependency check when the object is created. Maybe I'm missing something thought.

rcannood: Weren't you planning on moving all of the `read_h5ad_*` and so on functions into a separate package which could then be used by anndataR, mudata and spatialdataR?

lazappi: Ah, possibly... I had forgotten about that. Even if that does happen I think this would need to be revisited anyway. That package might just `import` **{rhdf5}** so this wouldn't be necessary there either.

rcannood (approved):
LGTM! (again)

PR #78, opened 2023-05-10
Update HDF5 tests

lazappi:
Update HDF5 tests so they (should) pass for CI etc.

- Correct the names in the "reading encoding" tests
- Adjust the expected values in the "Python -> R" roundtrip test
- Skip the "R -> Python" roundtrip test (needs writing of data frames which isn't implemented yet)


rcannood:
Good idea to skip tests until everything is implemented !

rcannood (approved):

PR #80, opened 2023-05-11
Issue-59-suggestions

rcannood:
* use `get_generator()` to get the correct generator
* combined the from_SCE and from_SCE_impl, since the code in from_SCE was small enough
* if no assays can be found, X should be NULL


rcannood:
Please ignore all of the base changes :zany_face:

mtmorgan (approved):

Issue #76, opened 2023-05-10
Make obs and var optional, make obs_names and var_names required


rcannood:
At the moment, we require that `obs` and `var` are passed to the InMemoryAnnData and HDF5AnnData, while other slots are optional.

However, looking at what happens when I create an empty AnnData in Python, it's actually the `obs_names` and the `var_names` that are required, and not `obs` and `var` -- I guess.

```python
import anndata as ad
adata = ad.AnnData(shape=[10, 20])
adata.write_h5ad("empty.h5ad")
```

```r
file <- rhdf5::H5Fopen("empty.h5ad")
rhdf5::h5ls(file)
```

      group   name       otype dclass dim
    0     / layers   H5I_GROUP           
    1     /    obs   H5I_GROUP           
    2  /obs _index H5I_DATASET STRING  10
    3     /   obsm   H5I_GROUP           
    4     /   obsp   H5I_GROUP           
    5     /    uns   H5I_GROUP           
    6     /    var   H5I_GROUP           
    7  /var _index H5I_DATASET STRING  20
    8     /   varm   H5I_GROUP           
    9     /   varp   H5I_GROUP

As such; I propose altering the code of `InMemoryAnnData` and `HDF5AnnData` require the `obs_names` and the `var_names` to be passed, not the `obs` and the `var`.

rcannood:
I agree this is very practical, but given #73 we can't add the obs_names to the obs as rownames or to the X as rownames because the obs_names might be integers instead of characters.

rcannood:
> What does Python anndata put in obs_names/var_names when you just set the shape?

This is what's shown in the original message, no?

lazappi:
> > What does Python anndata put in obs_names/var_names when you just set the shape?
> 
> This is what's shown in the original message, no?

Not entirely, but I could just check myself rather than asking you silly questions 😸

PR #81, opened 2023-05-11
make obs_names and var_names required

rcannood:
PR for addressing #76

* Makes `obs_names` and `var_names` required
* Makes `obs` and `var` optional
* Updates classes and test to reflect these changes accordingly


R/InMemoryAnnData.R
```diff
@@ -131,36 +123,36 @@ InMemoryAnnData <- R6::R6Class("InMemoryAnnData", # nolint
...
     #'   used to identify each row of `var` and to act as an index into
-    #'   the variable dimension of the AnnData object. For compatibility
-    #'   with *R* representations, `var_names` should be a character
-    #'   vector.
-    initialize = function(X = NULL, obs, var, obs_names = NULL, var_names = NULL, layers = NULL) {
-      # check obs and var first, because these objects are used by
+    #'   the variable dimension of the AnnData object. The length of
+    #'   the `var_names` defines the variable dimension of the AnnData
+    #'   object.
+    initialize = function(X = NULL, obs = NULL, var = NULL, obs_names, var_names, layers = NULL) {
```
lazappi, line 147: Can you move the compulsory arguments first?

R/AbstractAnnData.R
```diff
@@ -217,17 +220,10 @@ AbstractAnnData <- R6::R6Class("AbstractAnnData", # nolint
...
-      len <- switch(label,
-        obs = self$n_obs(),
-        var = self$n_vars()
-      )
-
-      if (!is.null(names) && length(names) != len) {
-        stop(wrap_message(
-          "length(", label, "_names) should be the same as ",
-          "nrow(", label, ")"
-        ))
```
lazappi, line 229: I'm guessing you removed this check because you might not know the dimensions yet? What happens if someone tries to set names of a different length later?

rcannood:
Update: in the last meeting we decided to go with `obs_names` and `var_names` after all :)

R/InMemoryAnnData.R
rcannood: Check!

R/AbstractAnnData.R
rcannood: Check! I added back the check, but changed it such that it will not trigger when the previous names are NULL, which should only be the case during object initialization.

PR #84, opened 2023-05-15
add remote to the description

rcannood:


rcannood:
@mtmorgan This seems to be working, except that the CI can't build the package from source because the hdf5 header files are probably not installed.

How about we disable the Mac OS X platform in the CI testing script so we can test out the rhdf5 branch?

PR #85, opened 2023-05-15
Update HDF5 writing functions and implement data frame writing

lazappi:
This implements writing of data frames to HDF5 files and generally updates the HDF5 writing functions

- Update `write_h5ad_encoding()` to work more reliably
- Rename `path_exists()` to `hdf5_path_exists()`
- `write_h5ad_element()` now deletes any existing elements before writing the new one. This should always be used instead of the specific writing functions.
- `write_h5ad_element()` now also passes arguments to writing functions using `...`, needed for data frames and has other potential uses.
- Implement `write_h5ad_data_frame()`
- Add tests for low level HDF5 writing functions
- General tidying

Still to do:

- Add `HDF5AnnData` setter tests (needs #74 to be merged)
- Turn on the reticulate round trip test. This needs the constructor to be rewritten to use `obs_names`/`var_names` instead of `obs`/`var` which felt like a big enough change for a separate PR.


rcannood (approved):
LGTM, though the proof of the pudding will be in the eating :)

PR #88, opened 2023-07-06
Add H5AD read/write helpers

lazappi:
Add user functions for reading/writing H5AD files to/from `SingleCellExperiment`/`Seurat` objects. This will likely be the main interface for most normal users.

Also adds functions for creating dummy data objects (based on the `dummy_data()` test helper).


lazappi:
@rcannood The CI here (and on other PRs) is failing on all platforms with an error that I think is coming from your **{anndata}** package (in the round-trip tests). I can't replicate it locally, though to work out what is going on. Any thoughts on how you want to handle this?

rcannood:
Sigh! That issue has been popping up everywhere. It's because scipy 1.11 introduced some breaking changes. The development version of reticulate (at rstudio/reticulate) has resolved the issue, but the changes still need to be released on CRAN. I'll add a workaround to the repo for now.

rcannood:
@lazappi The main branch contains a workaround by switching to the development version of reticulate for now. Can you merge the main branch into this branch?

R/HDF5-write.R
```diff
@@ -439,3 +441,50 @@ hdf5_path_exists <- function(file, target_path) {
...
+#' }
+#'
+#' # Write a Seurat as a H5AD
+#' h5ad_file <- tempfile(fileext = ".h5ad")
+#' if (requireNamespace("SeuratObject", quietly = TRUE)) {
+#'   seurat <- dummy_data(output = "Seurat")
+#'   write_h5ad(seurat, h5ad_file)
+#' }
+write_h5ad <- function(object, path) {
+  if (is(object, "SingleCellExperiment")) {
```
rcannood, line 471:
```suggestion
  if (inherits(object, "SingleCellExperiment")) {
```

R/HDF5-read.R
```diff
@@ -422,3 +422,35 @@ read_h5ad_collection <- function(file, name, column_order) {
...
+read_h5ad <- function(path, to = c("SingleCellExperiment", "Seurat")) {
+  to <- match.arg(to)
+
+  adata <- HDF5AnnData$new(path)
+
+  switch(to,
+    "SingleCellExperiment" = to_SingleCellExperiment(adata),
+    "Seurat" = to_Seurat(adata)
+  )
+}
```
rcannood, line 456: Can we make this function return an HDF5AnnData and an InMemoryAnnData as well?

```suggestion
read_h5ad <- function(path, to = c("HDF5AnnData", "InMemoryAnnData", "SingleCellExperiment", "Seurat")) {
  to <- match.arg(to)

  adata <- HDF5AnnData$new(path)

  switch(to,
    "SingleCellExperiment" = to_SingleCellExperiment(adata),
    "Seurat" = to_Seurat(adata),
    "HDF5AnnData" = adata,
    "InMemoryAnnData" = to_InMemory(adata)
  )
}
```

R/HDF5-write.R
```diff
@@ -439,3 +441,50 @@ hdf5_path_exists <- function(file, target_path) {
...
+    from_SingleCellExperiment(
+      object,
+      output_class = "HDF5AnnData",
+      file = path
+    )
+  } else if (is(object, "Seurat")) {
```
rcannood, line 477:
```suggestion
  } else if (inherits(object, "Seurat")) {
```

R/anndataR-package.R
```diff
@@ -1,5 +1,5 @@
 #' @keywords internal
-#' @importFrom methods as
+#' @importFrom methods as is
```
rcannood, line 2:
```suggestion
#' @importFrom methods as
```

R/HDF5-write.R
```diff
@@ -439,3 +441,50 @@ hdf5_path_exists <- function(file, target_path) {
...
+    from_Seurat(
+      object,
+      output_class = "HDF5AnnData",
+      file = path
+    )
+  } else {
```
rcannood, line 483: also allow writing object which inherit `AbstractAnnData`

```suggestion
  } else if (inherits(object, "AbstractAnnData")) {
    to_HDF5AnnData(object, path) # todo: implement
  }

**Review by rcannood — CHANGES_REQUESTED** [review]
These changes make sense to me, though I'd like to allow users to also return hdf5anndata objects instead of only Seurat or SCE. Is this ok with you?

**Comment by lazappi** [comment]
> These changes make sense to me, though I'd like to allow users to also return hdf5anndata objects instead of only Seurat or SCE. Is this ok with you?

I would prefer not to do this. Part of the reason for having these wrapper functions is to hide the `HDF5AnnData` interface object from most users. I can't see many use cases for interacting with it directly and I feel like anybody "advanced" enough to want to use it should be able to create the objects manually.

**Comment by rcannood** [comment]
Personally, I see no downsides to enabling this functionality, and again it's something that I actually need in order to use the package myself. Sure, I could do it the roundabout way without using `read_h5ad`, but I don't know why this should be the case when we might as well just add this functionality.

However, I can agree to not making the default. Would this be okay with you?

**Comment by lazappi** [comment]
I guess I can be convinced but I would maybe want to see what other people think. I think we just have a different philosophy/use case in mind. I know you have a use for it and could use it safely but you aren't a normal user and I worry about the average person misusing the interface objects if it's too easy to create them. Maybe I'm just paranoid though.

**Comment by rcannood** [comment]
I merged some changes into this PR.

@lazappi Do you have an idea why the devel branch fails when reading the obs in certain unit tests?

**Comment by rcannood** [comment]
Namely:

```
 ── Error ('test-HDF5-read.R:127:3'): reading H5AD as Seurat works ─────────���────
<subscriptOutOfBoundsError/error/condition>
Error in `element[["mask"]]`: subscript out of bounds
Backtrace:
    ▆
 1. ├─anndataR::read_h5ad(file, to = "Seurat") at test-HDF5-read.R:127:2
 2. │ └─anndataR::to_Seurat(adata) at anndataR/R/HDF5-read.R:452:2
 3. │   └─base::ncol(obj$obs) at anndataR/R/Seurat.R:42:2
 4. └─anndataR (local) `<fn>`() at anndataR/R/Seurat.R:42:2
 5.   └─anndataR:::read_h5ad_element(private$.h5obj, "/obs", include_index = FALSE) at anndataR/R/HDF5AnnData.R:43:8
 6.     └─anndataR (local) read_fun(file = file, name = name, version = version, ...) at anndataR/R/HDF5-read.R:67:2
 7.       └─anndataR:::read_h5ad_collection(file, name, column_order) at anndataR/R/HDF5-read.R:361:2
 8.         └─anndataR:::read_h5ad_element(...) at anndataR/R/HDF5-read.R:416:4
 9.           └─anndataR (local) read_fun(file = file, name = name, version = version, ...) at anndataR/R/HDF5-read.R:67:2
── Error ('test-HDF5AnnData.R:32:3'): reading obs works ────────────────────────
<subscriptOutOfBoundsError/error/condition>
Error in `element[["mask"]]`: subscript out of bounds
Backtrace:
    ▆
 1. └─anndataR (local) `<fn>`() at test-HDF5AnnData.R:32:2
 2.   └─anndataR:::read_h5ad_element(private$.h5obj, "/obs", include_index = FALSE) at anndataR/R/HDF5AnnData.R:43:8
 3.     └─anndataR (local) read_fun(file = file, name = name, version = version, ...) at anndataR/R/HDF5-read.R:67:2
 4.       └─anndataR:::read_h5ad_collection(file, name, column_order) at anndataR/R/HDF5-read.R:361:2
 5.         └─anndataR:::read_h5ad_element(...) at anndataR/R/HDF5-read.R:416:4
 6.           └─anndataR (local) read_fun(file = file, name = name, version = version, ...) at anndataR/R/HDF5-read.R:67:2

lazappi:
I think this might have to do with the changes to **{rhdf5}**. The nullable arrays are now read natively and you don't have to mess around with the masks yourself. I've fixed this in **{zellkonverter}** already so I can port that other here if you like.

rcannood:
I see! Is this related to #83 ?

lazappi:
Yes. Although I'm not sure if we are testing against the release or devel versions of Bioconductor packages (this change is only in devel).

rcannood:
The devel R CMD check uses the upcoming release of R, so it probably also has to use the development version of Bioconductor.

I created a PR (#90) which merges the changes into #83 into #88, which seems to have solved the issue. @lazappi Is it ok to merge #90?

lazappi:
I made one minor comment but otherwise yes.

rcannood (approved):

rcannood:
Thanks! #90 was merged. Can this PR also be merged?

lazappi:
Yeah, I think so

Issue #97, opened 2023-09-18
Automatically setup conda env


bschilder:
Add a function that auto-installs `miniconda` and then sets up a `conda` env to .
This would only be run when:
1) an `anndataR` function (or unit test) requires reticulate.
2) the conda env is not already present.

Could use either `reticulate` or `basilisk` for this.

bschilder:
Created new function: `setup_conda`

https://github.com/bschilder/anndataR/blob/rworkflows/R/setup_conda.R

lazappi:
This should use **{basilisk}** for Bioconductor compatibility. I'm not 100% sure we want to mess around with Python environments in this package though.

rcannood:
@lazappi Agreed :)

PR #105, opened 2023-09-18
Add rworkflows

bschilder:
# anndataR 0.99.0

## New features

- Add `rworkflows` CI.
- Update *lint.yaml* to use `actions/checkout@v4` (which has less issues).
- New function `setup_conda` automatically installs miniconda 
  and sets up conda env: #97
- Change version to Bioc-recommended devel version: 0.99.0


.github/workflows/pkgdown.yaml
rcannood, line 1: I think it's better to revert this change, because the site is being pushed to a different repository.

NEWS.md
```diff
@@ -1,5 +1,17 @@
...
+- Add `rworkflows` CI.
+- Update *lint.yaml* to use `actions/checkout@v4` (which has less issues).
+- New function `setup_conda` automatically installs miniconda 
+  and sets up conda env: #97
+- Change version to Bioc-recommended devel version: 0.99.0
+
 # anndataR 0.1.0
 
+## New features
+
```
rcannood, line 14: Version 0.1.0 hasn't been released yet, so we can just set the version to `0.99.0`. No need to add the "new features" to the changelog then, imo.

```suggestion
```

R/setup_conda.R
```diff
@@ -0,0 +1,29 @@
+#' Setup conda
+#' 
+#' Install miniconda and setup a conda env to use python dependencies.
+#' @inheritParams reticulate::install_miniconda
+#' @inheritParams reticulate::py_install
+#' @inheritDotParams reticulate::py_install
+#' @returns Null
+#' 
+#' @export
```
rcannood, line 9: I wouldn't export this function for now, at least until there is a reticulate backend. Also, we shouldn't create an Rd file for functions that aren't exported.

```suggestion
#' @noRd
```

man/setup_conda.Rd
rcannood, line 1: This should be removed (see earlier comment)

.github/workflows/R-CMD-check.yaml
```diff
@@ -1,51 +0,0 @@
...
-      - uses: r-lib/actions/setup-r-dependencies@v2
-        with:
-          extra-packages: any::rcmdcheck
-          needs: check
-
-      - name: Install Python dependencies for reticulate
-        run: |
-          reticulate::install_miniconda()
-          reticulate::py_install(c("anndata", "scanpy"), pip = TRUE)
-        shell: Rscript {0}
```
rcannood, line 47: Will `neurogenomics/rworkflows@master` also install these packages?

rcannood (changes requested):

NEWS.md
bschilder: Ok, I'll make the previous version 0.99.0 and my current version 0.99.1 to make the distinction that these were done separately.

.github/workflows/pkgdown.yaml
bschilder: Agreed! Though I'll also look into making rworkflows a bit more flexible for this use case.

.github/workflows/R-CMD-check.yaml
bschilder: Just to log our in-person discussion here, we talked about strategies for installing `miniconda`/`anndata`/`scanpy`. 

Here are some of the proposed options:
1. Trigger the `setup_conda` function whenever a `AnnDataR` function that requires `miniconda`/`anndata`/`scanpy` is used. We use this strategy for non-R dependencies in [`MAGMA.Celltyping`](https://github.com/neurogenomics/MAGMA_Celltyping), which requires the command line tool `MAGMA` for certain functions. The goal in that case is to remove any manual setup steps for the user. However, in the case of GHA, the only place that this would currently be triggered within `AnnDataR` is the unit tests, which @rcannood was opposed to. From a philosophical standpoint I have less of an issue with additional installs being triggered post-`AnnDataR` installation, but I do have some practical concerns that R might have some trouble using the `reticulate` conda env  right after it's been created. For this reason `basilisk` might be a better option, which switches between envs much more smoothly.
2. Add an optional step to rworkflows action to setup a conda env (I've actually been meaning to add this feature since its inception). Might be the best option tho, and shouldn't be too hard to implement (famous last words) 
3. On GHA, skip any tests that require `miniconda`/`anndata`/`scanpy`. Not a very good option, since it creates a huge blindspot during testing.

rcannood:
Can you merge the main branch into this PR?

bschilder:
> This PR contains a lot of changes which are not relevant to the rworkflows CI. I feel like this PR should only make changes to the .github folder (and maybe the README or the DESCRIPTION).

I was a bit too optimistic about my other PR and thought that would become "main" quite soon. Seems like those changes are going to take a while to break up and review, so i can change this PR back.

rcannood:
Closing this PR for now

PR #107, opened 2023-09-18
Extend roundtrip tests

rcannood:
Changes:

* Cover different matrix types: numeric_matrix, numeric_csparse, numeric_rsparse, numeric_matrix_with_nas, ...
* Cover different vector types: character, integer, factor, factor_ordered, logical, numeric, character_with_nas, ...
* Split up hdf5-reticulate roundtrip tests to more easily find out which data types are causing issues
* Incorporate regular `write_h5ad`-`read_h5ad` tests by @ivirshup 
* Dataset generator functions are no longer exported, as people shouldn't be using these, and we want to be able to change the interface in order to perform rigorous testing.

Note:

* Failing tests are disabled and will be re-enabled in a separate PR
* The dataset generator will probably be re-exported in a different PR


rcannood:
These tests are failing because some data types are causing issues.

rcannood:
@ivirshup Is it okay that I split up your roundtrip tests into different files?

I split them up because if something went wrong during a roundtrip test, it was hard to know which slot was causing the issue.

Also, while `test-roundtrip.R` tested both HDF5AnnData and InMemoryAnnData, I'm only testing HDF5AnnData here because using the InMemoryAnnData would still first read everything out with HDF5AnnData and then convert it to an InMemoryAnnData. It may be better to simply unit test the `to_InMemoryAnnData` function separately.

ivirshup:
> @ivirshup Is it okay that I split up your roundtrip tests into different files?

:+1:

> Also, while test-roundtrip.R tested both HDF5AnnData and InMemoryAnnData, I'm only testing HDF5AnnData here because using the InMemoryAnnData would still first read everything out with HDF5AnnData and then convert it to an InMemoryAnnData. It may be better to simply unit test the to_InMemoryAnnData function separately.

Up to you. I found this quite useful since it's shared assumptions across tests. Plus this is something that should be true in both cases.

If the InMemory case fails and the HDF5 one doesn't, then it was easy to tell the issue was with conversion of InMemory to HDF5

Issue #60, opened 2023-05-03
Implement `from_Seurat` and `to_Seurat`


rcannood:

bschilder:
## `from_Seurat ` strategy
- [ ] Only support single-assays when converting to "AbstractAnnData". 
- [ ] When a Seurat object has >1 assay, tell the user the Default assay is being used.
- [ ] Also allow the user to select which assay they want to use.
- [ ] Turn each matrix element within the selected assay into layers (`counts`, `data`, `scaled.data`).
- [ ] Set `X` as `counts` matrix element by default, as you can always normalise/manipulate the data from afterwards. However, allow the user the option of setting `X=NULL`

rcannood:
Sounds good, thanks!

csangara:
@bschilder @rcannood By the way, I will not have time to work on this anymore before the deadline. You can see what I've done so far in the `issue-60-from-seurat-obsm-varm` branch. I didn't get to test it very much yet so some parts will probably not work.

rcannood:
Fixed by #183

PR #118, opened 2023-09-19
`from_Seurat` updates

csangara:
From #60:
- Only support single-assays when converting to "AbstractAnnData".
- When a Seurat object has >1 assay, tell the user the Default assay is being used.
- Also allow the user to select which assay they want to use (with `assay` parameter)
- Turn each matrix element within the selected assay into layers (counts, data, scaled.data).
- Set X as counts matrix element by default, but also allow the user the option of setting X=NULL or X=c('counts', 'data', 'scale.data')

Still to do:
- obsm / varm


rcannood:
@csangara Can you run `styler::style_pkg()`?

You can also add `Rscript -e 'styler::styke_pkg()'` to your `.git/hooks/pre-commit` to automate this step for you.

R/Seurat.R
```diff
@@ -149,33 +170,31 @@ from_Seurat <- function(seurat_obj, output_class = c("InMemoryAnnData", "HDF5Ann
...
+    # Cannot compare other values with NULL
+    X <- "none"
+  }
+
+  # Add the remaining non-empty slots as layers
+  slots <- c("counts", "data", "scale.data")
+  slots <- slots[slots != X]
+
+  for (slot in slots) {
+    if (!all(dim(GetAssayData(seurat_obj, slot = slot)) == 0)) {
```
rcannood, line 194: GetAssayData is not imported

```suggestion
    if (!all(dim(SeuratObject::GetAssayData(seurat_obj, slot = slot)) == 0)) {
```

rcannood (changes requested):

csangara:
Ok, thanks for the tip! I keep running it manually but I guess I forgot in the most recent commit.

rcannood (approved):

PR #122, opened 2023-09-19
Make actions run on PRs on forks

ivirshup:


.github/workflows/R-CMD-check.yaml
```diff
@@ -1,6 +1,9 @@
 # Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
 # Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
-on: push
+on:
+  push:
+  pull_request:
+    branches: [main, master]
```
ivirshup, line 6: Could be reasonable to just run on all pull requests, regardless of target?

```suggestion
```

rcannood (approved):

PR #135, opened 2023-11-29
Pass all R CMD and BiocChecks

bschilder:
After merging the main upstream branch, I made corrections / enhancements throughout to get anndataR passing all R CMD and Bioc checks.

Here's the summary from the updated NEWS file:
# anndataR 0.99.1

## New features

* Harmonise all references to `h5ad_file` (`file`,`h5ad_path`,`ad`)
* Harmonise all references to `ad` (`adata`)
* Use `@returns` in Roxygen notes to allow multi-line notes.
* Export helper function `dummy_data`
  - Makes examples much less verbose and more consistent.
  - Include "obsm","varm","obsp","varp" and "layers" wherever possible.
* *Seurat.R*
  - Add/extract `DimRed` elements whenever available.
  - Add unit tests for `from_Seurat`.
* New functions
  - `graphs_to_anndata`:  When converting Seurat --> AnnData, 
    store `obsp` and `varp` anndata slots from `graphs`. Does some inference 
    about whether each matrix should go into the `obsp` vs. `varp` slot based 
    on whether they have the same dimensions as the number of rows. 
  - `anndata_to_graphs`: When converting AnnData --> Seurat, 
    store `obsp` and `varp` anndata slots in `graphs`. 
* *DimReduc.R*
  - New functions for converting Seurat's DimReduc objects.
  

## Bug fixes

* *DESCRIPTION*:
  - Add "email" tag to `Authors`.
  - Add `BiocViews`
* Fix `BiocCheck` WARNINGS:
  - `Empty or missing \value sections found in man pages.`
* Fix `BiocCheck` NOTES:
  - Make all vignettes use `BiocStyle`
  - `Update R version dependency from 3.2.0 to 4.3.0.`
  - `'sessionInfo' not found in vignette(s)`
  - `Avoid the use of 'paste' in condition signals` 
  - `Avoid 'cat' and 'print' outside of 'show' methods`
  - `Consider adding runnable examples to man pages that document exported objects.`
  - Added helper function `messager` to avoid notes about pasting inside messages.
    Can also easily turn off verbosity by setting the env var "ANNDATAR_VERBOSE".
* Fix all unit tests given updates.

<hr>
Wasn't sure about the versioning so I can change that if you want.


DESCRIPTION
```diff
@@ -57,11 +57,14 @@ Description: Bring the power and flexibility of AnnData to the R
 License: MIT + file LICENSE
 URL: https://scverse.org/anndataR, https://github.com/scverse/anndataR
 BugReports: https://github.com/scverse/anndataR/issues
+BioViews:
+  Software
 Depends: 
-    R (>= 3.2.0)
+    R (>= 4.3.0)
```
rcannood, line 63: Why bump the dependency on R to >=4.3? Do you have reason to believe this package won't work with older versions of R?

R/AbstractAnnData.R
```diff
@@ -77,8 +78,9 @@ AbstractAnnData <- R6::R6Class("AbstractAnnData", # nolint
     #'   computationally expensive.
     #' @param ... Optional arguments to print method.
     print = function(...) {
-      cat("AnnData object with n_obs \u00D7 n_vars = ", self$n_obs(), " \u00D7 ", self$n_vars(), "\n", sep = "")
-
+      messager("AnnData object with n_obs \u00D7 n_var =",
+               self$n_obs(), "\u00D7 ",
+               self$n_var(), "\n")
```
rcannood, line 83: You accidentally removed some spacing in the message. Also, you added trailing whitespaces which linter claims is superfluous.

R/DimReduc.R
```diff
@@ -0,0 +1,192 @@
...
+#'  there are conventional names for the loadings slots
+#' @inheritDotParams SeuratObject::CreateDimReducObject
+#' @returns A named list of \link[SeuratObject]{CreateDimReducObject} objects.
+#'
+#' @export
+#' @examples
+#' adata <- dummy_data("InMemoryAnnData")
+#' drl <- to_DimReduc(obj=adata)
+#' drl
+to_DimReduc <- function(obj = NULL,
```
rcannood, line 26: Why is this function here? This should be a part of the `to_Seurat` and not a standalone function. Or did you reach a consensus to do this differently?

R/dummy_data.R
```diff
@@ -0,0 +1,303 @@
...
+#' @param ... Additional arguments passed to subfunctions
+#'  (determined by \code{output}).
+#' 
+#'
+#' @returns Object containing the generated dataset as defined by `output`
+#'
+#' @export
+#' @examples
+#' dummy <- dummy_data("InMemoryAnnData")
+dummy_data <- function(output = c(
```
rcannood, line 22: This function was removed.

rcannood (changes requested):
Hi @bschilder !

I apologize, but this PR is way too big. It seems like you're trying to address multiple things at the same time, and makes changes in almost all of the files that are currently in AnnData.

Looking at the first couple of pages, it seems like the proposals you make will:

* Make changes to the Seurat converter
* Add functions for converting AnnData to Seurat graphs and dimreds -- this should be incorporated into the Seurat converter itself
* Add back dummy data while it was previously removed
* Apply formatting to the code (but also introduce several linting issues yourself)
* Add a 'massager' function
* Rename `n_vars()` to `n_var`: Why? This is called `n_vars` in Python anndata

Since this PR makes changes to so many files in the package, accepting it would mean that everybody else has to deal with a lot of merge conflicts.

I think that many of these changes were made for an older version of the repository, since this PR introduces back some functions that were previously removed from AnnData (i.e. the dummy_data functions).

However, I see a lot of useful contributions in this PR.

Can you make separate PRs where each PR contains just one "topic" that you want to address? For example, where you make changes to the DESCRIPTION and the vignettes, one where you make changes to X, one where you make changes to Y. This will make it easier to review the proposed changes, or see whether some proposed changes causes problems for somebody else.

Thanks @bschilder !

DESCRIPTION
bschilder: It's a Bioconductor requirement, Bioc releases are synced with specific versions of R

rcannood: Where is this documented? I don't see any Bioc package doing this.

SummarizedExperiment has `R (>= 4.0.0)`: https://code.bioconductor.org/browse/SummarizedExperiment/blob/RELEASE_3_18/DESCRIPTION
BiocGenerics has `R (>= 4.0.0)`: https://code.bioconductor.org/browse/BiocGenerics/blob/RELEASE_3_18/DESCRIPTION
S4Vectors has `R (>= 4.0.0)`: https://code.bioconductor.org/browse/S4Vectors/blob/RELEASE_3_18/DESCRIPTION

R/AbstractAnnData.R
```diff
@@ -111,15 +113,15 @@ AbstractAnnData <- R6::R6Class("AbstractAnnData", # nolint
...
+        self$n_var()
       )
     },
     #' @description Number of observations in the AnnData object.
     n_obs = function() {
       length(self$obs_names)
     },
     #' @description Number of variables in the AnnData object.
-    n_vars = function() {
+    n_var = function() {
```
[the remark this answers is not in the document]
bschilder: i thought it was a typo, and was trying to keep with anndata conventions. 
but just checked the anndata docs and you're indeed right, ill switch it back

https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.n_vars.html

R/DimReduc.R
bschilder: The Seurat functions are getting super long, and I find it quite helpful to break things into testable subfunctions after a certain threshold.

Or do you simply mean why is it in "this file" and not "Seurat.R"?

R/messager.R
```diff
@@ -0,0 +1,20 @@
...
+#'  and meet Bioconductor requirements that dictate the message 
+#'  must first be stored to a variable before passing to \link[base]{message}. 
+#'  
+#' @param v Whether to print messages or not.
+#' @param parallel Whether to enable message print when wrapped 
+#' in parallelised functions.
+#' @inheritParams base::paste
+#' @return Null 
+#' @keywords internal 
+messager <- function(...,
```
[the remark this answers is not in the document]
bschilder: Of what, specifically?

R/DimReduc.R
rcannood: Splitting up functionality into multiple functions makes sense. However, I don't think that the `to_DimReduc` is a function that users should be able to access themselves. Additionally, I think all of the seurat functions should be prefixed with something so we can recognize it's seurat-related. For example, a lot of helper functions for reading and writing h5ad files are stored in `read_h5ad_helpers.R` and `write_h5ad_helpers.R` and are start with `read_h5ad_` and `write_h5ad_` respectively.

R/dummy_data.R
bschilder: Ok, was there a strong reason behind that?

You'll notice we have a lot of redundant and verbose code just to create the same objects over again.  many of the tests and examples are mostly setup steps just get the objects needed to test a one-line function, which I feel kind of distracts from the purpose of the documentation in those cases. 

It also means if wan to change how our examples are set up, we have to go through every location and fix them independently, which leaves us prone to missing some instances during updates and thus having inconsistencies. 

Of course, there will occasionally be scenarios where we don't to customise the example data further, or fully expose the object creation process.

bschilder:
> I think that many of these changes were made for an older version of the repository, since this PR introduces back some functions that were previously removed from anndataR (i.e. the dummy_data functions).

I moved those from the test folder into the main R/ folder bc i found them really helpful in examples. But maybe that's something to discuss in more detail at our next meeting.

Were there any other files you noticed that seemed old? I fetched and pulled from the upstream main branch, but it's entirely possible something didn't go right during this process.

R/DimReduc.R
bschilder: Got it, thanks for clarifying. 

Can def rename the function, but we seem to be using a number of different conventions for grouping functions in the code so I was sure which would be most appropriate here.

For example, we could group by the prefix "Seurat_....." making these function "Seurat_to"
or the prefix could be "to_..." (as it is currently) making it "to_Seurat".
The latter has Seurat as the suffix instead of the prefix, but it's arguable more intuitive to read.

Depending on what we decide, this would make the function in question something like:
- "to_Seurat_DimReduc"
- "to_DimReduc_Seurat"
- "Seurat_DimReduc_to"

Maybe another things to add to the discussion next week.

DESCRIPTION
bschilder: There's multiple levels of version control enforcement:
- The DESCRIPTION file
-`BiocManager` 

If you try to install Bioc 3.18 on a version of R that's too old, you get an error.

See the docs here:
https://bioconductor.org/about/release-announcements/
> Each Bioconductor release is designed to work with a specific version of R. 

This file tracks the R-Bioc versions:
https://bioconductor.org/config.yaml

bschilder:
We've decided to break this PR into multiple smaller PRs to make merging more manageable.

PR #138, opened 2023-12-02
re-enable h5ad categorical tests

rcannood:
This tests results in a warning.


rcannood:
Update: Apparently it doesn't. Maybe it was already fixed, or it was a temporary issue on my system only.

PR #139, opened 2023-12-02
re-enable test writing h5ad from SCE

rcannood:
This tests results in one or more warnings.


rcannood:
Update: Apparently it doesn't. Maybe it was already fixed, or it was a temporary issue on my system only.

PR #150, opened 2023-12-02
re-enable uns roundtrip test with nested list

rcannood:
Probably related to all other testing PRs.

The nested list should probably also be a bit bigger (different matrix and data frame formats and deeply nested listsI


rcannood:
"list" isn't part of the default arguments while it should be

PR #152, opened 2023-12-04
export `generate_dataset()` function

### LINKED ISSUE #136: Move dummy data functions back to main package
It would be helpful to have the dummy data functions accessible from the main package, not just the tests. Mainly, this would be useful for examples so that they are shorter, more consistent and easier to write.

rcannood:
Changes:

* Re-exports the dataset generator function.
* Reworked the arguments of the dataset generator function
* Renamed internal functions
* Fixed print method

Closes #136.


R/generate_dataset.R
```diff
@@ -138,77 +204,96 @@ generate_dataset_as_list <- function(
...
-#' Generate a dummy dataset as a SingleCellExperiment object
-#'
-#' @param ... Parameters passed to `generate_dataset_as_list`
+#' @param list Output of `.generate_dataset_as_list()`
 #'
 #' @return SingleCellExperiment containing the generated data
 #'
 #' @noRd
-generate_dataset_as_sce <- function(...) {
+.generate_dataset_as_sce <- function(list) {
```
lazappi, line 214: I would probably rename the `list` variable to something like `dataset_list`, just to avoid confusion with the `list()` function

R/generate_dataset.R
```diff
@@ -4,43 +4,108 @@
...
+      "numeric_dense_with_nas", "numeric_csparse_with_nas", "numeric_rsparse_with_nas", "integer_matrix",
+      "integer_dense", "integer_csparse", "integer_rsparse", "integer_matrix_with_nas", "integer_dense_with_nas",
+      "integer_csparse_with_nas", "integer_rsparse_with_nas"
+    ),
+    varp_types = c(
+      "numeric_matrix", "numeric_dense", "numeric_csparse", "numeric_rsparse", "numeric_matrix_with_nas",
+      "numeric_dense_with_nas", "numeric_csparse_with_nas", "numeric_rsparse_with_nas", "integer_matrix",
+      "integer_dense", "integer_csparse", "integer_rsparse", "integer_matrix_with_nas", "integer_dense_with_nas",
+      "integer_csparse_with_nas", "integer_rsparse_with_nas"
+    ),
```
lazappi, line 74: I like that all these are included but maybe we could make an easy way to generate a reduced list for examples?

R/generate_dataset.R
```diff
@@ -138,77 +204,96 @@ generate_dataset_as_list <- function(
...
       "Creating a Seurat requires the 'SeuratObject' package to be installed"
     )
   }
 
-  dummy <- generate_dataset_as_list(...)
-
-  X <- t(dummy$layers[["integer_csparse"]])
-  colnames(X) <- dummy$obs_names
-  rownames(X) <- dummy$var_names
+  X <- t(list$layers[["integer_csparse"]])
```
lazappi, line 255: Should this be `list$X` in case `integer_csparse` wasn't generated? Same with the ones below.

R/get_generator.R
```diff
@@ -10,6 +12,8 @@ list_generators <- function() {
 #'
 #' @param class Name of the AnnData class. Must be one of `"HDF5AnnData"`
 #' or `"InMemoryAnnData"`.
+#'
+#' @noRd
 get_generator <- function(class = c("HDF5AnnData", "InMemoryAnnData")) {
   # TODO: also support directly passing the correct class?
```
lazappi, line 18: I'm not quite sure what the purpose of this is. Is `generator` here different to the dataset generators?

tests/testthat/test-generate_dataset.R
```diff
@@ -19,3 +19,15 @@ test_that("generating dummy Seurat works", {
...
+args1 <- formals(generate_dataset)
+args2 <- formals(.generate_dataset_as_list)
+
+# If any of these tests are failing, the arguments of generate_dataset()
+# need to be updated to include the new generator types.
+for (arg in intersect(names(args1), names(args2))) {
+  test_that(paste0("generate_dataset(): argument '", arg, "' has correct default value"), {
+    expect_equal(eval(args1[[arg]]), eval(args2[[arg]]))
+  })
+}
```
lazappi, line 33: Ah, I guess it's for this. If it's not reused somewhere else I would just list things directly here but this also works.

lazappi (approved):
I've made some comments but I think it's pretty good. I would also update the examples to use this but maybe not everyone wants to do that.

R/get_generator.R
rcannood: Yes -- the generator function is used by the `from_Seurat` and `from_SingleCellExperiment` to create the correct type of AnnData object:

```r
from_SingleCellExperiment <- function(sce, output_class = c("InMemory", "HDF5AnnData"), ...) { # nolint
  stopifnot(
    inherits(sce, "SingleCellExperiment")
  )

  # fetch generator
  generator <- get_generator(output_class)

  # ... create obs, var, ...

  generator$new(
    X = x,
    obs = obs,
    var = var,
    obs_names = obs_names,
    var_names = var_names,
    layers = layers,
    ...
  )
}
```

I agree, however, that the term `generator` is now somewhat confusing, given that there is a `generate_dataset()` function. Fortunately, the `get_generator` function is an internal one, so we can rename it to whatever we like :)

R/generate_dataset.R
rcannood: True -- though I think the Seurat conversion should be revised in light of Seuratv5. Maybe we can just pass all layers to Seurat without needing to be selective.

R/generate_dataset.R
rcannood: I see. Not sure how we could go about this, other than perhaps:

```r
generate_dummy_dataset <- function(n_obs, n_vars) {
  generate_dataset(
    n_obs = n_obs,
    n_vars = n_vars,
    obs_types = c(...), # small set of types
    ...
  )
}
```

rcannood: Aha, wait! We could turn it around. We could set the defaults `generate_dataset()` to something simple, and then create a helper function in the tests which contains all of the types.

R/get_generator.R
lazappi: Yeah, it might be better to have a more specific name. `object_generator` or `converter` maybe?

R/generate_dataset.R
rcannood: I implemented a solution for this in 206f11a64c9bfed7fb6aa51d5684a2a2447ccfe9

Issue #121, opened 2023-09-19
Previously written fields are not cleaned up


ivirshup:
```r
library(anndataR)
library(Matrix)

a = AnnData(obs_names=as.character(1:10), var_names=as.character(1:10), layer=list(foo=matrix(1, 10, 10)))
write_h5ad(a, "test.h5ad")

b = AnnData(obs_names=as.character(1:100), var_names=as.character(1:10), X=matrix(0, 100, 10))
write_h5ad(b, "test.h5ad")
```

```bash
$ h5ls -r test.h5ad
/                        Group
/X                       Dataset {100, 10}
/layers                  Group
/layers/foo              Dataset {10, 10}
/obs                     Group
/obs/_index              Dataset {100}
/obsm                    Group
/obsp                    Group
/uns                     Group
/var                     Group
/var/_index              Dataset {10}
/varm                    Group
/varp                    Group
```

`layers/foo` is still in the file, though `b` did not have that entry. `layers/foo` is also not of the correct shape, making the file invalid.

lazappi:
Example of the current behaviour. You get an error now but it's not a very helpful one.

``` r
library(anndataR)

a <- AnnData(X = matrix(0, 10, 20), layers = list(foo = matrix(1, 10, 20)))
write_h5ad(a, "test.h5ad")

b <- AnnData(X = matrix(0, 10, 20), layers = list(bar = matrix(2, 10, 20)))
write_h5ad(b, "test.h5ad")
#> Error in H5File.open(filename, mode, file_create_pl, file_access_pl): HDF5-API Errors:
#>     error #000: ../../../src/H5F.c in H5Fcreate(): line 349: unable to create file
#>         class: HDF5
#>         major: File accessibility
#>         minor: Unable to open file
#> 
#>     error #001: ../../../src/H5Fint.c in H5F_open(): line 1725: unable to open file
#>         class: HDF5
#>         major: File accessibility
#>         minor: Unable to open file
#> 
#>     error #002: ../../../src/H5FD.c in H5FD_open(): line 722: open failed
#>         class: HDF5
#>         major: Virtual File Layer
#>         minor: Unable to initialize object
#> 
#>     error #003: ../../../src/H5FDsec2.c in H5FD__sec2_open(): line 351: unable to open file: name = '/tmp/RtmpuLZhqk/reprex-8dca2f482d3c-gem-drake/test.h5ad', errno = 17, error message = 'File exists', flags = 15, o_flags = c2
#>         class: HDF5
[...]

PR #156, opened 2023-12-05
Clean file when writing H5AD

lazappi:
Address #121 where old fields are not removed from files when writing to an existing H5AD file. This PR currently solves this by adding an `overwrite` argument to `write_h5ad()`. Possibly needs more work depending on what interfaces we want exposed to the user.


rcannood (commented):
I'm not sure whether it's currently possible to trigger this condition: https://github.com/scverse/anndataR/blob/ee9ee49a36a7e3e2b8ea8faece61d4efea11b2e3/R/HDF5AnnData.R#L283

Either the file doesn't exist yet, or the constructor will fail if you try to open an H5AD file and change stuff at the same time.

Could you take a look? :bow:

PR #157, opened 2023-12-05
Tell RStudio to strip trailing whitespace

lazappi:
Set `StripTrailingWhitespace: Yes` in `anndataR.Rproj` so that RStudio strips trailing whitespace when files are saved (and **{lintr}** can stop complaining 😸).

Alternatively, we can remove `anndata.Rproj` from the repo which is preferred/required by Bioconductor anyway.


rcannood (approved):
Both solutions are fine by me! Feel free to remove it at a later point if you like.

PR #160, opened 2023-12-05
allow h5 files that have already been opened with rhdf5::H5Fopen

rcannood:


lazappi (changes requested):
Could we add some tests for this? I suspect some of the reading/writing functions will fail.

rcannood:
This functionality is no longer necessary due to #169

PR #170, opened 2024-07-03
Generate Python based anndata testfiles

LouiseDck:
Instead of relying on round-trip tests, we also want to be able to read in all possible Python arrays, matrices and dataframes.


rcannood:
> Instead of relying on round-trip tests, we also want to be able to read in all possible Python arrays, matrices and dataframes.

I think we should still do round-trip tests, but the difference is that our current round-trip tests currently:

* use R functions to generate data
* use CRAN anndata to move R data into Python with reticulate
* use Python anndata to write to data to disk
* use anndataR to read data from disk
* compare original data to resulting data

(the reverse round-trip is also tested)

While this PR will enable tests:

* use reticulate to call Python functions which generate data
* use Python anndata to write data to disk
* use anndataR to read data from disk
* compare original data to resulting data

What I'm currently wondering about is how to compare the original data in Python to the resulting data in R. Do we introduce a simple JSON format to store which assertions to make (i.e. let the python function write out which assertions R should be making to check whether the anndata was created correctly)? Or do we let the Python functions generate data in a predictable manner so that we can reproduce the same results in R?

LouiseDck:
> I think we should still do round-trip tests, but the difference is that our current round-trip tests currently:

    use R functions to generate data
    use CRAN anndata to move R data into Python with reticulate
    use Python anndata to write to data to disk
    use anndataR to read data from disk
    compare original data to resulting data

I agree. 

> What I'm currently wondering about is how to compare the original data in Python to the resulting data in R. Do we introduce a simple JSON format to store which assertions to make (i.e. let the python function write out which assertions R should be making to check whether the anndata was created correctly)? Or do we let the Python functions generate data in a predictable manner so that we can reproduce the same results in R?

After discussion we settled on generating the same dataset in both Python and R, and checking if there are differences when reading in both datasets, and reporting differences using h5diffs.

LouiseDck:
For documentation purposes: the python anndata generator functions have moved to a separate package: [dummy-anndata](https://github.com/data-intuitive/dummy-anndata)

LouiseDck:
Superseded by #207 and [dummy-anndata](https://github.com/data-intuitive/dummy-anndata)

PR #171, opened 2024-07-04
Make rownames part of obs and var

rcannood:
These changes have been ported from #166 and #169.

Rownames have been moved from a separate obs_names and var_names to being the rownames of obs and var, since it simplifies the way the data is read to and from an hdf5 file.


LouiseDck (approved):
I have no issues with the way you refactored `obs_names` and `var_names`.

If I recall correctly though, the reason for not doing it this way in the first place was the following:
rownames in R can only be character vectors, which is an assumption that does not (or will not) apply to the Python anndata version. I believe that the impact of this is relatively minor, however I do feel it is necessary to be documented  somewhere?
(This is in general: we need to have some documentation delineating the differences and what is and isn't supported)

rcannood:
Good idea! I was going to add it to the roxygen docs, but found it a bit weird to document differences between R and Python while the Python AnnData also only allows string obs_names and var_names for now. I'll create a separate issue.

lazappi:
> I have no issues with the way you refactored `obs_names` and `var_names`.
> 
> If I recall correctly though, the reason for not doing it this way in the first place was the following: rownames in R can only be character vectors, which is an assumption that does not (or will not) apply to the Python anndata version. I believe that the impact of this is relatively minor, however I do feel it is necessary to be documented somewhere? (This is in general: we need to have some documentation delineating the differences and what is and isn't supported)

You are right. I had forgotten but this was (at least partly) why we changed it before. I would probably make a special man page and/or vignette for things we know are different from Python.

rcannood:
> I would probably make a special man page and/or vignette for things we know are different from Python.

Good point! I added it to #172.

PR #183, opened 2024-09-19
Refactor from/to Seurat conversion

rcannood:
* Redesigned the from_Seurat / to_Seurat conversion functions


rcannood:
Note: Could use https://satijalab.github.io/seurat-object/reference/UpdateSeuratObject.html to update a Seurat <v5 object to v5 in `from_Seurat()`.

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
-  # create seurat object
-  if (ncol(var_) > 0) {
-    # don't add var metadata if the data frame does not contain any columns
-    x_assay <- SeuratObject::AddMetaData(x_assay, metadata = var_)
+  # trackstatus: class=Seurat, feature=get_obsp, status=wip
+  for (i in seq_along(graph_mapping)) {
+    graph_name <- names(graph_mapping)[[i]]
+    graph <- graph_mapping[[i]]
+    if (!is.character(graph) || length(graph) != 1) {
+      stop("each graph must be a character vector of length 1")
```
lazappi, line 205: Maybe this is clearer:

```suggestion
      stop("item ", graph_name, "in graph_mapping is not a character vector of length 1")
```

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
-  # add layers
-  # trackstatus: class=Seurat, feature=get_layers, status=wip
-  # TODO: should values be passed to counts or to data?
-  for (key in obj$layers_keys()) {
-    layer_ <- t(obj$layers[[key]])
-    dimnames(layer_) <- list(var_names_, obs_names_)
-    seurat_obj[[key]] <- SeuratObject::CreateAssayObject(counts = layer_)
+  # trackstatus: class=Seurat, feature=get_uns, status=done
+  # trackstatus: class=Seurat, feature=get_varp, status=done
+  obj@misc <- list()
```
lazappi, line 217: Can probably also avoid the `@` here

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+      stop("misc_mapping must be a named list with one or two elements")
+    }
+    expected_slots <- c("X", "layers", "obs", "obsm", "obsp", "var", "varm", "varp", "uns")
+    if (!misc[[1]] %in% expected_slots) {
+      stop(paste0(
+        "The first element of each item in misc_mapping must be one of: ",
+        paste0("'", expected_slots, "'", collapse = ", ")
+      ))
+    }
+    misc_data <- adata[[misc[[1]]]]
```
lazappi, line 231: Is it possible to get the names without getting everything in the slot? Just because for `H5ADAnnData` this will drag everything in that slot into memory (I think).

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+  # trackstatus: class=Seurat, feature=get_varp, status=done
+  obj@misc <- list()
+  for (i in seq_along(misc_mapping)) {
+    misc_name <- names(misc_mapping)[[i]]
+    misc <- misc_mapping[[i]]
+    if (!is.character(misc) || length(misc) <= 0 || length(misc) > 2) {
+      stop("misc_mapping must be a named list with one or two elements")
+    }
+    expected_slots <- c("X", "layers", "obs", "obsm", "obsp", "var", "varm", "varp", "uns")
+    if (!misc[[1]] %in% expected_slots) {
```
lazappi, line 225: The code would be a bit easier to follow if there was something like:

```r
misc_slot <- misc[[1]]
misc_key <- misc[[2]]
```

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
   }
 
-  names
+  if (!.to_seurat_is_atomic_character(layer_name)) {
+    stop("layer_name must be the name of one of the layers or NULL")
+  }
+
+  if (!layer_name %in% names(adata$layers)) {
+    stop("layer_name must be the name of one of the layers or NULL")
+  }
```
lazappi, line 267: These could have different messages to make it clearer what the problem is.

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+
+  return(Matrix::t(adata$layers[[layer_name]]))
+}
+
+.to_seurat_get_matrix <- function(adata, layer_name) {
+  if (is.null(layer_name)) {
+    return(Matrix::t(adata$X))
+  }
```
lazappi, line 275: Do you need to do this or can you assume there will always be a name? It might depend what the guessing function do which I haven't looked at yet.

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+  if (!is.null(varm_loadings) && !.to_seurat_is_atomic_character(varm_loadings)) {
+    stop("varm_loadings must be a character scalar or NULL")
+  }
+  embed <- adata$obsm[[obsm_embedding]]
+
+  if (is.null(embed)) {
+    warning(paste0("The reduction ", obsm_embedding, " is not present in adata$obsm, skipping"))
+    return(NULL)
+  }
```
lazappi, line 304: I think other missing things are an error rather than a warning?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+    assay = assay_name,
+    global = TRUE
+  )
+
+  if (!is.null(loadings)) {
+    rownames(loadings) <- adata$var_names
+    args$loadings <- loadings
+  }
+
+  do.call(SeuratObject::CreateDimReducObject, args)
```
lazappi, line 330: Is there a reason to do it this way rather than just calling the function with arguments?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+to_Seurat_guess_layers <- function(adata) { # nolint
+  if (!inherits(adata, "AbstractAnnData")) {
+    stop("adata must be an object inheriting from AbstractAnnData")
+  }
+
+  layers <- list()
+
+  if (!is.null(adata$X)) {
+    layers["counts"] <- list(NULL)
+  }
```
lazappi, line 342: Ah, I guess you use `NULL` to mean `adata$X` rather than `adata$layers["X"]`. I missed that before.

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#' @export
+#'
+#' @rdname to_Seurat
 #' @examples
 #' ad <- AnnData(
 #'   X = matrix(1:5, 3L, 5L),
 #'   obs = data.frame(row.names = LETTERS[1:3], cell = 1:3),
 #'   var = data.frame(row.names = letters[1:5], gene = 1:5)
 #' )
 #' to_Seurat(ad)
```
lazappi, line 81: It would be nice to have some examples showing how the mapping works.

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#' @section Layer mapping:
+#'
+#' A named list to map AnnData layers to Seurat layers. Each item in the list must be a character vector of length 1.
+#' The `$X` key maps to the `X` slot.
+#'
+#' Example: `layer_mapping = list(counts = "counts", "$X" = "data", foo = "bar")`.
+#'
+#' If `NULL`, the internal function `from_Seurat_guess_layers` will be used to guess the layer mapping as follows:
+#'
+#' * All Seurat layers are copied by name.
```
lazappi, line 450: I assume to `adata$layers` and `adata$X` is empty?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' @section Obsm mapping:
+#'
+#' A named list to map Seurat reductions to AnnData `obsm`. Each item in the list must be a named list with keys 'key',
+#' 'obsm', and 'varm'.
+#'
+#' Example: `obsm_mapping = list(pca = list(key = "PC_", obsm = "X_pca", varm = "PCs"))`.
```
lazappi, line 457: I'll see below but it should probably be possible for some of these to be missing

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' If `NULL`, the internal function `from_Seurat_guess_obsms` will be used to guess the obsm mapping as follows:
+#'
+#' * All Seurat reductions are prefixed with `X_` and copied to AnnData `obsm`.
+#'
+#' @section Varm mapping:
+#'
+#' A named list to map Seurat PCA loadings to AnnData `varm`.
```
lazappi, line 465: Is `obsm_mapping` also copying things to `varm`? Do we need both?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' Example: `varm_mapping = list(pca = "PCs")`.
+#'
+#' If `NULL`, the internal function `from_Seurat_guess_varms` will be used to guess the varm mapping as follows:
+#'
+#' * The name of the PCA loadings is copied by name.
```
lazappi, line 471: Only PCA or any other reduction that has loadings?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' @section Obsp mapping:
```
lazappi, line 473: I would not capitalise obsp and maybe format it as code

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' A named list to map Seurat miscellaneous data to AnnData `varp`. Each item in the list must be a character
+#' vector of length 3 or less. The first element must be `"misc"`. The second element is the name of the data in the
+#' corresponding slot. The third element is used to subset the data even further.
+#'
+#' Example: `varp_mapping = list(foo = c("misc", "foo"))`.
+#'
+#' If `NULL`, the internal function `from_Seurat_guess_varps` will be used to guess the varp mapping as follows:
+#'
+#' * No data is mapped to `varp`.
```
lazappi, line 493: This bit I don't understand. Why have a first element if it's always the same and what does the third element do?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#' obj <- CreateSeuratObject(counts = counts)
+#' obj <- NormalizeData(obj)
+#' obj <- FindVariableFeatures(obj)
+#' obj <- ScaleData(obj)
+#' obj <- RunPCA(obj, npcs = 10L)
+#' obj <- FindNeighbors(obj)
+#' obj <- RunUMAP(obj, dims = 1:10)
+#' from_Seurat(obj)
 # nolint start: object_name_linter
 from_Seurat <- function(
```
lazappi, line 523: Maybe these functions should be called something like `anndata_from_seurat()` just to make it a bit clearer?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
     # nolint end: object_name_linter
     seurat_obj,
     output_class = c("InMemoryAnnData", "HDF5AnnData"),
-    assay = NULL,
-    X = "counts",
+    assay_name = "RNA",
```
lazappi, line 527: This could be the name of the first assay by default in case it's something else

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
-  if (!is.null(assay)) {
-    if (!assay %in% names(seurat_obj@assays)) {
-      stop("'assay' must be NULL or one of: ", paste0("'", names(seurat_obj@assays), "'", collapse = ", "))
-    }
-    assay_name <- assay
-  } else {
-    assay_name <- SeuratObject::DefaultAssay(seurat_obj)
+  if (is.null(seurat_assay) || !inherits(seurat_assay, "Assay5")) {
+    stop(paste0("Assay '", assay_name, "' is not a valid Seurat v5 assay"))
+  }
```
lazappi, line 543: I think this is checking two things. I would probably have one check for the object version (with a prompt to upgrade it) and a second check later for the assay name.

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
 
-    # If Seurat object contains multiple assays, notify user the Default one is used
-    if (length(names(seurat_obj@assays)) > 1) {
-      message(
-        "There are ", length(names(seurat_obj@assays)), " assays in the Seurat object; using the default assay ('",
-        assay_name, "'). You can use the `assay` parameter to select a specific assay."
-      )
-    }
```
lazappi, line 153: I think this is nicer behaviour, I would put this back

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+  }
+  if (is.null(uns_mapping)) {
+    uns_mapping <- from_Seurat_guess_uns(seurat_obj)
   }
 
-  # get obs
+  # fetch obs
   # trackstatus: class=Seurat, feature=set_obs_names, status=done
   # trackstatus: class=Seurat, feature=set_obs, status=done
   obs <- seurat_obj@meta.data
```
lazappi, line 567: I'm pretty sure `seurat_obj[[]]` will get this but would need to double check

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+  var <- seurat_assay@meta.data
 
   # use generator to create new AnnData object
   generator <- get_anndata_constructor(output_class)
-  ad <- generator$new(
+  adata <- generator$new(
     obs = obs,
     var = var,
     ...
   )
```
lazappi, line 580: Maybe it's better to wait until the end of the function to create this? Otherwise you could end up with incomplete files for `H5ADAnnData`.

lazappi (changes requested):
I made a bunch of comments, most of them are fairly minor. I haven't tried running anything yet though so possible I missed things.

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+  colnames(obj) <- obs_names
+  rownames(obj) <- var_names
+
+  # copy other layers
+  for (i in seq_along(layer_mapping)) {
+    from <- layer_mapping[[i]]
+    to <- names(layer_mapping)[[i]]
+    if (to == "counts" || to == "data") {
+      next
+    }
```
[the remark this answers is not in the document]
LouiseDck: I don't mind this control flow

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+    SeuratObject::LayerData(obj, assay = assay_name, layer = to) <- adata$layers[[from]]
+  }
+
+  # copy reductions
+  # trackstatus: class=Seurat, feature=get_obsm, status=wip
+  # trackstatus: class=Seurat, feature=get_varm, status=wip
+  if (!is.null(reduction_mapping)) {
+    if (!is.list(reduction_mapping) || (length(reduction_mapping) > 0 && is.null(names(reduction_mapping)))) {
+      stop("reduction_mapping must be a named list")
     }
```
LouiseDck, line 175: Lots of checking, but it is still possible that not all elements of the list are named. If e.g. only one of the elements is named, the unnamed reductions get mapped to an "X" reduction in Seurat, and overwrite each other.

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
-    } else {
-      mat <- Matrix::sparseMatrix(
-        i = integer(0),
-        p = c(0L),
-        x = integer(0),
-        dims = c(obj$n_vars(), obj$n_obs())
+      if (
+        !is.list(reduction) || is.null(names(reduction)) ||
+          !all(names(reduction) %in% c("key", "obsm", "varm")) ||
+          !all(c("key", "obsm", "varm") %in% names(reduction))
```
LouiseDck, line 183: I have an issue with requiring the "varm" name in the list: for many dimensionality reductions you do not get loadings in "varm", so requiring it seems a bit harsh?

R/Seurat.R
LouiseDck: Out of curiosity, why do you want to avoid the `@`?

R/Seurat.R
LouiseDck, line 450: I want it emphasized that this means that by default, `adata$X` will be `NULL`, and that if you want to prevent this you should prevent a mapping.

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' If `NULL`, the internal function `from_Seurat_guess_obsps` will be used to guess the obsp mapping as follows:
+#'
+#' * All Seurat graphs are copied to `obsp` by name.
+#'
+#' @section Varp mapping:
+#'
+#' A named list to map Seurat miscellaneous data to AnnData `varp`. Each item in the list must be a character
+#' vector of length 3 or less. The first element must be `"misc"`. The second element is the name of the data in the
+#' corresponding slot. The third element is used to subset the data even further.
```
LouiseDck, line 487: Maybe this is my inexperience with Seurat showing, but what do you mean by `used to subset the data further`?

R/Seurat.R
LouiseDck: I disagree, as we only aim to convert from & to anndata.
However, if we want to emphasize this, I would prefer `anndata_to_seurat()` and `seurat_to_anndata()`

R/Seurat.R
LouiseDck: I agree, if you want to pick a default assay to convert it should not be `RNA` by default, but the `DefaultAssay`

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+
+    key1 <- obsm[[1]]
+    key2 <- obsm[[2]]
+
+    if (key1 == "reductions") {
+      adata$obsm[[obsm_name]] <- seurat_obj@reductions[[key2]]@cell.embeddings
+    } else if (key1 == "misc") {
+      adata$obsm[[obsm_name]] <- seurat_obj@misc[[key2]]
+    }
+  }
```
LouiseDck, line 614: I might overlook this, but there's no need to specifiy "varm" in the mapping, right? In that case, I think the documentation of the function needs an update.

DESCRIPTION
```diff
@@ -65,12 +65,12 @@ Imports:
     R6
 Suggests:
     anndata,
-    BiocStyle,
```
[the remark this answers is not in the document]
LouiseDck: cf #169 but we should indeed discuss this

LouiseDck (changes requested):

DESCRIPTION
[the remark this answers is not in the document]
rcannood: Removed changes that were not related to Seurat and moved the discussion to #191 !

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#' @param adata An AnnData object to be converted
+#' @param assay_name Name of the assay to be created
+#' @param layer_mapping A named list to map AnnData layers to Seurat layers. See section "Layer mapping" for more
+#'   details.
+#' @param reduction_mapping A named list to map AnnData reductions to Seurat reductions. Each item in the list must be a
+#'   named list with keys 'key', 'obsm', and 'varm'. See section "Reduction mapping" for more details.
+#' @param graph_mapping A named list to map AnnData graphs to Seurat graphs. Each item in the list must be a character
+#'   vector of length 1. See section "Graph mapping" for more details.
+#' @param misc_mapping A named list to map miscellaneous data to the names of the data in the Seurat object. See section
+#'   "Miscellaneous mapping" for more details.
```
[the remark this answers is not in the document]
rcannood: At the moment I named the arguments based on what the object that ends up being created. In `to_Seurat` we're deciding on what needs to be copied into the Seurat object, which is why the argument names are named after seurat object slots.

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' @section Layer mapping:
+#'
+#' A named list to map AnnData layers to Seurat layers. Each item in the list must be a character vector of length 1.
+#' A value of `NULL` corresponds to the `X` slot.
+#'
+#' Example: `layer_mapping = list(counts = "counts", data = NULL, foo = "bar")`.
```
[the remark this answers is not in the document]
rcannood: Fixed!

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' If `NULL`, the internal function `to_Seurat_guess_layers` will be used to guess the layer mapping as follows:
```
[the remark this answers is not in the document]
rcannood: Originally each of the functions were being exported but this resulted in a lot of clutter in the package documentation, which is why they are currently internal functions. Would you prefer to export them?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' * If `adata$X` is defined, we assume this the `counts`.
+#' * Other layers are copied by name.
+#'
+#' @section Reduction mapping:
+#'
+#' A named list to map AnnData `obsm` and `varm` to Seurat reductions. Each item in the list must be a named list
```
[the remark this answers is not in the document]
rcannood: Hmm. Or what about "AnnData `$obsm` and `$varm`"?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#' Example: `reduction_mapping = list(pca = list(key = "PC_", obsm = "X_pca", varm = "PCs"))`.
+#'
+#' If `NULL`, the internal function `to_Seurat_guess_reductions` will be used to guess the reduction mapping as follows:
+#'
+#' * All obsm starting with `X_` are copied by name.
+#'
+#' @section Graph mapping:
+#'
+#' A named list mapping graph names to the names of the graphs in the AnnData object. Each item in the list must be a
+#' character vector of length 1.
```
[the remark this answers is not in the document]
rcannood: Fixed!

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' * An obsp named `connectivities` will be mapped to `nn`.
+#' * Other graphs starting with `connectivities_` are stripped of the prefix and copied by name.
 #'
-#' `to_Seurat()` converts an AnnData object to a Seurat object.
+#' @section Miscellaneous mapping:
 #'
-#' @param obj An AnnData object
+#' A named list mapping miscellaneous data to the names of the data in the AnnData object. Each item in the list must be
+#' a named list with one or two elements. The first element must be one of: 'X', 'layers', 'obs', 'obsm', 'obsp', 'var',
```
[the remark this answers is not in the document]
rcannood: Fixed!

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#' 'varm', 'varp', 'uns'. The second element is the name of the data in the corresponding slot. If the second element is
+#' not present, the data in the first element will be used.
```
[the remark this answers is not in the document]
rcannood: I tried to clarify -- is this better?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+#'
+#' Example: `misc_mapping = list(uns = "uns", varp_neighbors = c("varp", "neighbors"))`.
+#'
+#' If `NULL`, the internal function `to_Seurat_guess_misc` will be used to guess the miscellaneous mapping as follows:
+#'
+#' * If `adata$obsp` is defined, this is mapped to a misc value named `obsp`.
+#' * If `adata$uns` is defined, this is mapped to a misc value named `uns`.
```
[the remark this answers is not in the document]
rcannood: :+1:!

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+# nolint start: object_name_linter
+to_Seurat <- function(
+    adata,
+    assay_name = "RNA",
+    layer_mapping = NULL,
+    reduction_mapping = NULL,
+    graph_mapping = NULL,
+    misc_mapping = NULL) {
+  # nolint end: object_name_linter
   requireNamespace("SeuratObject")
```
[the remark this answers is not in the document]
rcannood: :+1:!

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
-  var_ <- obj$var
-  rownames(var_) <- var_names_
+  # store obs and var names
+  obs_names <- adata$obs_names[]
+  var_names <- adata$var_names[]
+
+  # check seurat layers
+  if (is.null(names(layer_mapping))) {
+    names(layer_mapping) <- layer_mapping
+  }
```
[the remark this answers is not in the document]
rcannood: I'd say it's okay with using `layer_mapping = c("counts", "data")` as a synonym for `c(counts = "counts", data = "data")`.

R/Seurat.R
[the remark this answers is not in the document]
rcannood: I changed 

```r
    if (to == "counts" || to == "data") {
      next
    }
```

to 

```r
    if (!to %in% c("counts", "data")) {
      SeuratObject::LayerData(obj, assay = assay_name, layer = to) <- adata$layers[[from]]
    }
```

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+  obj <- SeuratObject::CreateSeuratObject(
+    meta.data = adata$obs,
+    assay = assay_name,
+    counts = counts,
+    data = data
+  )
+
+  # trackstatus: class=Seurat, feature=get_var, status=done
+  if (!is.null(adata$var)) {
+    obj@assays[[assay_name]] <- SeuratObject::AddMetaData(
```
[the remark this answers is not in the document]
rcannood: Let me know if you'd prefer something else :relaxed:

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+      obj@assays[[assay_name]],
+      metadata = adata$var
+    )
+  }
+
+  # make sure obs and var names are set properly
+  # trackstatus: class=Seurat, feature=get_obs_names, status=done
+  # trackstatus: class=Seurat, feature=get_var_names, status=done
+  colnames(obj) <- obs_names
+  rownames(obj) <- var_names
```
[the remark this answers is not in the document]
rcannood: Maybe. Seurat might have done something with those dimnames by then...

R/Seurat.R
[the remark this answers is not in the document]
rcannood: :+1:

R/Seurat.R
rcannood: I believe @ is usually used for accessing internal data structures which end users should not be messing around with. 

I'm not sure whether there is always an easy alternative. Let me know if you find one :)

R/Seurat.R
rcannood: Good point. Probably not, unless we 

```r
if (slot_name == "obs") adata$obs_keys()
elif (slot_name == "var") adata$var_keys()
elif ...
```

R/Seurat.R
rcannood: done!

R/Seurat.R
rcannood: true!

R/Seurat.R
rcannood: well you need to have some way of allowing the user to use `$X` as the seurat counts of data, so by using `NULL` the `$X` will be used.

R/Seurat.R
rcannood: True. This was a leftover from when the defaults of to_seurat and from_seurat just had a bunch of defaults without checking whether these exist in the original anndata / seurat object.

R/Seurat.R
rcannood: fixed!

R/Seurat.R
rcannood: :+1:

R/Seurat.R
rcannood: true!

R/Seurat.R
rcannood: Yes, I'll clarify.

R/Seurat.R
rcannood: Sorry, the documentation is no longer correct, I'll fix.

R/Seurat.R
rcannood: Sorry, the `obsm_mapping` documentation was wrong -- is fixed now.

R/Seurat.R
rcannood: Tried to clarify the docs

R/Seurat.R
rcannood: I'd be fine with:

* `anndata_to_seurat` & `seurat_to_anndata`
* `to_seurat` & `from_seurat`
* or keep it as is, namely `to_Seurat` and `from_Seurat`

R/Seurat.R
rcannood: true -- `obj@active.assay`?

R/Seurat.R
rcannood: I'll add it to the error message!

R/Seurat.R
rcannood: Fixed!

R/Seurat.R
rcannood: True. Otoh, it would make the whole operation a whole lot less memory efficient.

I could add a tryCatch which removes the file on error. WDYT?

R/Seurat.R
```diff
@@ -1,205 +1,815 @@
...
+  for (i in seq_along(obsm_mapping)) {
+    obsm <- obsm_mapping[[i]]
+    obsm_name <- names(obsm_mapping)[[i]]
+
+    if (!is.character(obsm) || length(obsm) != 2) {
+      stop("each obsm_mapping must be a character vector of length 2")
+    }
+
+    key1 <- obsm[[1]]
+    key2 <- obsm[[2]]
```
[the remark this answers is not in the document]
rcannood: +1

R/Seurat.R
rcannood: To not make it too complicated, can we assume that reduction_mapping is always a named list?

R/Seurat.R
rcannood: +1!

R/Seurat.R
rcannood: +1!

R/Seurat.R
rcannood: Phrasing was improved

R/Seurat.R
rcannood: not sure what you mean -- could you clarify?

R/Seurat.R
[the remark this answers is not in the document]
lazappi: Not necessarily, pros and cons. But if they were exported then the documentation should go there.

R/Seurat.R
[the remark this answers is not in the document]
lazappi: Yeah, maybe. I just think it's easy to get mixed up between different objects/arguments so it helps to be as clear as possible. I don't really mind how exactly, as long as it's consistent.

R/Seurat.R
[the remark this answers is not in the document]
lazappi: Yep, as long as it matches what it does in practice. What it says now is different to what I understood before.

R/Seurat.R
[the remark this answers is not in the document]
lazappi: Ok. If there is a reason for it then it's fine.

R/Seurat.R
lazappi: In theory yes (or a case statement might be nicer) but I think at the moment `adata$obs_keys()` etc. read everything anyway so not much point unless those are changed. Maybe don't worry about it for this PR.

R/Seurat.R
lazappi: Maybe, if there's a nice way to do it. The way we write stuff at the moment is a bit risky in general so maybe it's more a design flaw in `H5ADAnnData`...

R/Seurat.R
rcannood: Good point, still need to change this

R/Seurat.R
rcannood: I changed the default to NULL. If set to NULL, `SeuratObject::DefaultAssay(seurat_obj)` will be used.

R/Seurat.R
rcannood: Ok, let me know ^^

R/Seurat.R
```diff
@@ -1,205 +1,830 @@
...
+  if (!inherits(seurat_obj, "Seurat")) {
+    stop("The provided object must be a Seurat object")
+  }
+
+  varm_mapping <- list()
+
+  if ("pca" %in% names(seurat_obj@reductions)) {
+    # Check if the dimreduc was calculated by the selected assay
+    reduction <- seurat_obj@reductions[["pca"]]
+    if (reduction@assay.used == assay_name) {
```
lazappi, line 776: I can't find a better way to get the assay used so probably have to use `@` here

R/Seurat.R
```diff
@@ -1,205 +1,830 @@
...
+}
+
+from_Seurat_guess_obsps <- function(seurat_obj, assay_name) { # nolint
+  if (!inherits(seurat_obj, "Seurat")) {
+    stop("The provided object must be a Seurat object")
+  }
+
+  obsp_mapping <- list()
+
+  for (graph_name in names(seurat_obj@graphs)) {
```
lazappi, line 791:
```suggestion
  for (graph_name in SeuratObject::Graphs(seurat_obj)) {
```

R/Seurat.R
```diff
@@ -1,205 +1,830 @@
...
+    graph <- seurat_obj@graphs[[graph_name]]
```
lazappi, line 792:
```suggestion
    graph <- seurat_obj[[graph_name]]
```

lazappi (changes requested):
Mostly adding suggestions for accessing stuff from a Seurat object. I didn't comment on every line but I think there should be an example for each slot.

R/Seurat.R
LouiseDck: Yes -- My comment was wrongly placed and you already solved part of it :tada:

R/Seurat.R
```diff
@@ -1,205 +1,830 @@
...
 #' @param output_class Name of the AnnData class. Must be one of `"HDF5AnnData"` or `"InMemoryAnnData"`.
-#' @param assay Assay to be converted. If NULL, `DefaultAssay()` is used.
-#' @param X Which of 'counts', 'data', or 'scale.data' will be used for X. By default, 'counts' will be used (if it is
-#'   not empty), followed by 'data', then 'scale.data'. The remaining non-empty slots will be stored in different
-#'   layers.
+#' @param assay_name The name of the assay to be converted. If `NULL`, the default assay will be used.
+#' @param layer_mapping A named list mapping layer names to the names of the layers in the Seurat object. Each item in
+#' the list must be a character vector of length 1. See section "Layer mapping" for more details.
+#' @param obsm_mapping A named list mapping reduction names to the names of the reductions in the Seurat object.
+#' Each item in the list must be a named list with keys 'key', 'obsm', and 'varm'. See section "Reduction mapping" for
```
LouiseDck, line 423: This seems wrong -- it also does not correspond to the "obsm mapping" section.

R/check_requires.R
```diff
@@ -0,0 +1,34 @@
...
+
+  if (any(!is_available)) {
+    missing <- requires[!is_available]
+    missing_str <- paste0("'", paste(missing, collapse = "', '"), "'") # nolint object_usage_linter
+    cli_abort(
+      c(
+        "{what} requires the {.pkg {missing}} package{?s}",
+        "i" = paste(
+          "To continue, install {cli::qty(missing)}{?it/them} using",
+          "{.code install.packages(c({missing_str}))}"
```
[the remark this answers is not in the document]
rcannood: > These instructions won't work for Bioconductor packages.

Fixed!

> updating all messages/warning/errors to use {cli} if we are going to depend on it anyway?

I'd like that, but let's do this in a separate PR :sweat_smile:

R/Seurat.R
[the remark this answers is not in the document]
rcannood: Tried to make the notations more consistent!

R/Seurat.R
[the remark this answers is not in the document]
rcannood: `GetAssayData` and `SetAssayData` returns the data of the assay, not the assay itself.

I can use `obj[[assay_name]]` instead of `obj@assays[[assay_name]]` but I don't see the added benefit. If anything, `obj[[assay_name]]`  allows me to accidentally interact with data that is not actually an assay.

Upon looking closer, I see that SeuratObject provides [documentation of the Seurat class](https://satijalab.github.io/seurat-object/reference/Seurat-class.html) for slots like `@assays`, `@graphs`, ..., so I'd say using these slots is actually ok.

R/Seurat.R
rcannood: Doesn't work as a replacement for this line of code.

```
SeuratObject::Misc(obj) <- list()
Error in `Misc<-.Seurat`(`*tmp*`, value = list()) : 
  argument "slot" is missing, with no default
```

In this case, we could **assume** that `obj@misc` is an empty list to begin with.

R/Seurat.R
rcannood: Thanks for noticing, has been fixed!

R/Seurat.R
rcannood: It's possible to expose `ad$to_Seurat()` and not `to_Seurat(ad)`, but we'll need to expose `from_Seurat(obj)` as this functionality is otherwise not accessible. This is the status quo ^^

R/Seurat.R
```diff
@@ -1,205 +1,830 @@
...
   output_class <- match.arg(output_class)
 
   stopifnot(inherits(seurat_obj, "Seurat"))
 
-  if (!is.null(X)) {
-    if (!X %in% c("counts", "data", "scale.data")) {
-      stop("X must be NULL or one of: 'counts', 'data', 'scale.data'")
-    }
+  if (is.null(assay_name)) {
+    assay_name <- SeuratObject::DefaultAssay(seurat_obj)
```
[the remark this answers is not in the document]
rcannood: This would bypass the `requiredNamespace("SeuratObject")` ^^

I'll make a note in the docs.

R/Seurat.R
```diff
@@ -1,205 +1,830 @@
...
 
-  # If a specific assay is selected, use it
-  if (!is.null(assay)) {
-    if (!assay %in% names(seurat_obj@assays)) {
-      stop("'assay' must be NULL or one of: ", paste0("'", names(seurat_obj@assays), "'", collapse = ", "))
-    }
-    assay_name <- assay
-  } else {
-    assay_name <- SeuratObject::DefaultAssay(seurat_obj)
+  seurat_assay <- seurat_obj@assays[[assay_name]]
```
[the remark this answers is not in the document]
rcannood: `GetAssayData()` returns the data layer of the assay, not the assay object itself.

R/Seurat.R
rcannood: [`@meta.data`](https://satijalab.github.io/seurat-object/reference/Seurat-class.html) is a documented slot of the Seurat -- I like the more explicit nature of this notation.

R/Seurat.R
```diff
@@ -1,205 +1,830 @@
...
-      ad$layers[[slot]] <- Matrix::t(assay_data)
+    if (!is.character(obsp) || length(obsp) < 2 || length(obsp) > 3) {
+      stop("each obsp_mapping must be a character vector of length 2 or 3")
     }
+
+    key1 <- obsp[[1]]
+    key2 <- obsp[[2]]
+
+    if (key1 == "graphs") {
+      adata$obsp[[obsp_name]] <- as(seurat_obj@graphs[[key2]], "sparseMatrix")
```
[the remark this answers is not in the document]
rcannood: Then lets stick to using @graphs, as this is more explicit. I don't want strange errors to occur because a user accidentally provided the name of a DR as a graph name.

R/Seurat.R
```diff
@@ -1,205 +1,830 @@
...
+
+  seurat_assay <- seurat_obj@assays[[assay_name]]
+
+  if (!inherits(seurat_assay, "Assay5")) {
+    stop("The provided object must be a Seurat v5 assay")
+  }
+
+  layers_mapping <- list()
+
+  for (layer_name in names(seurat_assay@layers)) {
```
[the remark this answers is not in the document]
rcannood: The code was used to use these functions where possible :)

R/Seurat.R
```diff
@@ -1,205 +1,830 @@
...
+from_Seurat_guess_obsms <- function(seurat_obj, assay_name) { # nolint
+  if (!inherits(seurat_obj, "Seurat")) {
+    stop("The provided object must be a Seurat object")
+  }
+
+  obsm_mapping <- list()
+
+  for (reduction_name in names(seurat_obj@reductions)) {
+    # Check if the dimreduc was calculated by the selected assay
+    reduction <- seurat_obj@reductions[[reduction_name]]
```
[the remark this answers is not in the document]
rcannood: @reductions is kept as a safety precaution

R/Seurat.R
rcannood: `@graphs` is kept as a precaution

R/Seurat.R
rcannood: I changed it to:

```

  tryCatch(
    {
      generator$new(
        obs = obs,
        var = var,
        ...
      )
      # ... add more data ...
    },
    error = function(e) {
      if (output_class == "HDF5AnnData") {
        on.exit(cleanup_HDF5AnnData(...))
      }
      stop(e)
    }
  )
```

wdyt?

rcannood:
Phew. I think I addressed most of the comments. 

* Wherever SeuratObject had functions for accessing certain data, these functions were used (e.g. changed `names(obj@layers)` for `SeuratObject::Layers(obj)`). However, I didn't change code when the alternative was less specific (e.g. changing `obj@assays[[assay_name]]` for `obj[[assay_name]]`) or readable (e.g. changing `obj@meta.data` for `obj[[]]`) to avoid accidental bugs.
* Tried to clean up with `cleanup_HDF5AnnData` when an error occurs while converting to HDF5AnnData. @lazappi Can you take a look?
* Changing all `stop()`, `warning()`, `cat()`, etc.. to use **{cli}** sounds great, but let's do that in a separate PR

R/SingleCellExperiment.R
```diff
@@ -168,11 +171,19 @@ from_SingleCellExperiment <- function(
...
+      X = x,
+      obs = obs,
+      var = var,
+      layers = layers,
+      ...
+    ),
+    error = function(e) {
+      if (output_class == "HDF5AnnData") {
+        on.exit(cleanup_HDF5AnnData(...))
+      }
```
lazappi, line 185: Should this be outside the error part so the cleanup is still run if one of the later steps fails?

lazappi (approved):
I think I'm happy with this now

R/SingleCellExperiment.R
rcannood: I think there is no code after this call in the function. The SCE converter is simply a bit limited at this stage

PR #190, opened 2024-11-05
Add Zarr support

### LINKED ISSUE #91: Implement zarr backend

keller-mark:
Fixes #91 

These changes are from both me and @Artur-man 

The main public-facing changes here are:
- The `ZarrAnnData` class
- `read_zarr` and `write_zarr` top-level functions
- Support for `from_Seurat(output_class="ZarrAnnData")`
- Support for `from_SingleCellExperiment(output_class="ZarrAnnData")`

Internally:
- `read_zarr_helpers.R` is the zarr analog of `read_h5ad_helpers.R`
- `write_zarr_helpers.R` is the zarr analog of `write_h5ad_helpers.R`
- Test fixtures within `inst/extdata/example.zarr` (this makes the diff noisy, apologies)
- Lots of tests:
  - `test-Zarr-read.R` (35 new tests)
  - `test-Zarr-write.R` (70)
  - `test-ZarrAnnData.R` (26)
  - `test-h5ad-zarr.R` (17)
 
A number of these functions generate warnings in the R console that are intended to be followed up on to improve the code (and should probably be resolved as end users may not appreciate them), but the tests still pass despite these warnings.

Known things that are not implemented here:
- support for `recarrays`
- usage of `mode = c("r", "r+", "a", "w", "w-", "x")` parameter value


R/read_zarr.R
```diff
@@ -0,0 +1,47 @@
...
+  to <- match.arg(to)
+
+  adata <- ZarrAnnData$new(path)
+
+  fun <- switch(to,
+    "SingleCellExperiment" = to_SingleCellExperiment,
+    "Seurat" = to_Seurat,
+    "InMemoryAnnData" = to_InMemoryAnnData,
+    "ZarrAnnData" = NULL
+  )
```
rcannood, line 40: Should add HDF5AnnData as an option here as well. Which means that we should probably store this information elsewhere to avoid code duplication

```suggestion
read_zarr <- function(
    path,
    to = c("InMemoryAnnData", "HDF5AnnData", "SingleCellExperiment", "Seurat", "ZarrAnnData"),
    ...) {
  to <- match.arg(to)

  adata <- ZarrAnnData$new(path)

  fun <- switch(to,
    "SingleCellExperiment" = to_SingleCellExperiment,
    "Seurat" = to_Seurat,
    "InMemoryAnnData" = to_InMemoryAnnData,
    "HDF5AnnData" = to_HDF5AnnData,
    "ZarrAnnData" = NULL
  )

File: R/read_zarr_helpers.R [hunk 2]
```diff
@@ -0,0 +1,510 @@
...
+#' @return A named list with names type and version
+#'
+#' @noRd
+read_zarr_encoding <- function(store, name, stop_on_error = TRUE) {
+  # Path can be to array or group
+  g <- pizzarr::zarr_open(store, path = name)
+  attrs <- g$get_attrs()$to_list()
+
+  if (!all(c("encoding-type", "encoding-version") %in% names(attrs))) {
+    path <- name
```
rcannood, line 17: where are a lot of linting issues in this file -- could you run `lintr::lint_package()` and fix any issues that pop up?

inst/extdata/example2.zarr/.zgroup
rcannood, line 1: This file should probably be removed

rcannood (changes requested):
Fantastic work @keller-mark and @Artur-man !

I went through the PR for a first time and left some minor comments. I will review the code by running it a couple of times next :)

inst/extdata/example2.zarr/.zgroup
Artur-man: done

R/read_zarr_helpers.R
Artur-man: working on it now ...

Artur-man:
@rcannood @keller-mark should we add `pizzarr` only under `Suggests` in `DESCRIPTION` and no need to add a `Remotes` list here ? 

```
Suggests:
    ...
    ...
    pizzarr
Remotes:
    keller-mark/pizzarr
```

R/read_zarr_helpers.R
Artur-man: done ... did a full `lint_package` check and corrected some R check issues too!

R/read_zarr.R
Artur-man: done ... mixed this with the latest commit! https://github.com/scverse/anndataR/pull/190/commits/d192e68ed312f212476a5490c973619f08e7c8de

keller-mark:
Good point, at minimum we should add a note to the Github README here https://github.com/scverse/anndataR/tree/main?tab=readme-ov-file#installation

Artur-man:
Ah, this is already added here, nevermind then only `Suggests` should work. 
https://github.com/keller-mark/anndataR/blob/d192e68ed312f212476a5490c973619f08e7c8de/R/ZarrAnnData.R#L240-L246

lazappi:
Not a proper review of the code but I wanted to raise that we plan to submit **{anndataR}** to Bioconductor (or maybe CRAN) at some point (hopefully soonish). This means we won't be able to depend on GitHub only packages. Do you have plans to submit **{pizzarr}** somewhere or could you rewrite this to use one of the public Zarr packages?

Artur-man:
@lazappi thanks for the notification, `pizzarr` could be submitted to CRAN soon. https://github.com/keller-mark/pizzarr/pull/93

Artur-man:
@keller-mark any opinions on this ? perhaps we can make Rarr/pizzarr optional with `options()` depending on the speed. 

Regardless it would be good to have zarr support for AnnDataR. We are currently using `reticulate`/`basilisk` env in `SpatialData` and causes problems already (too many dependancies, multiple env etc.). https://github.com/HelenaLC/SpatialData

That being said, any dependancy should either be in CRAN or BioC. If there are missing utilities in `Rarr` we can also contact the maintainer and/or send PRs there (I did some, which I can continue too).

inst/extdata/example.zarr/.zattrs
lazappi, line 1: Can we please zip or something the example to make it a single file? I don't think we want to be tracking ~400 extra files in `git`. If we can avoid needing an example file that might be even better but I'm happy to keep it if it's bundled in some way.

Artur-man: should be easy to zip and unzip at tmpdir, will do

Artur-man: done

Artur-man:
I have quickly checked if `pizzarr` utilities could be replaced with https://github.com/grimbough/Rarr, unfortunately there exists a set of limitations to the BioC native package, which 
- [x] opening zarr stores and creating groups, see https://github.com/Huber-group-EMBL/Rarr/pull/18
- [x] read/write boolean arrays, see https://github.com/Huber-group-EMBL/Rarr/pull/59
- [x] fixing errors when reading from some character arrays, see https://github.com/grimbough/Rarr/issues/20
   
Will be in touch to see if these are resolved in the future, otherwise no zarr R package is currently both in CRAN/BioC and functionally complete yet.

Artur-man:
There is some progress in the Rarr package, would you guys like a clean PR (since there were so many updates since) or continuing here is fine ?

lazappi:
> There is some progress in the Rarr package, would you guys like a clean PR (since there were so many updates since) or continuing here is fine ?

Probably whatever is easiest for you and @keller-mark/whatever makes the PR easiest to understand. There have been a lot of changes to the package since this was opened so we would need to make sure those are included here.

I saw that **{Rarr}** is planning to have Zarr v3 support for the next release so I think that makes sense in terms of which backend package to use.

Artur-man:
I adapted the changes, was fairly easy. I will then continue here and refer back to Hugo if needed again.

lazappi:
I'm going to make this a draft for now, just to help with our organisation. Please let us know when it's ready to review.

Artur-man:
Guys, I am close .... I will add a new `example.zarr` here, however lets add a yml file to allow creating quick environments, so the same files can be generated quickly later, and also ... would you like a separate `example_zarr.py` file ? 

https://github.com/scverse/anndataR/blob/5ffd79a05bbc4a99194b89a6432777cf2f1760a5/inst/scripts/example_h5ad.py#L1-L7

lazappi:
We tried to cover as many cases as possible in the example dataset so I would probably modify the script to also output a (tarred/zipped) Zarr as well as an H5AD (and maybe rename it to something like `create_example.py`) rather than creating a new one.

I noted down the versions just in case it became important later but maybe it makes sense to store them in a separate file. If you are regenerating the dataset I would probably also update the environment. I think there is also a version number in the script that should be bumped and a changelog that should be updated.

R/Rarr_utils.R
```diff
@@ -0,0 +1,164 @@
+#' create_zarr_group
+#'
+#' create zarr groups
+#'
+#' @param store the location of (zarr) store
+#' @param name name of the group
+#' @param version zarr version
+#' @export
+create_zarr_group <- function(store, name, version = "v2"){
+  split.name <- strsplit(name, split = "\\/")[[1]]
```
Bisaloo: It'll be slightly faster if regexes are not used:

```suggestion
  split.name <- strsplit(name, split = "/", fixed = TRUE)[[1]]
```

R/Rarr_utils.R
```diff
@@ -0,0 +1,164 @@
...
+  if(length(split.name) > 1){
+    split.name <- vapply(seq_len(length(split.name)),
```
Bisaloo:
```suggestion
    split.name <- vapply(seq_along(split.name),
```

lazappi:
@Artur-man We have updated the CI. You will need to merge the `devel` branch here to get the new version.

R/write_zarr_helpers.R
```diff
@@ -0,0 +1,553 @@
...
+                                    name,
+                                    compression,
+                                    version = "0.2.0",
+                                    overwrite = FALSE) {
+
+  if (!is.null(dim(value))) {
+    dims <- dim(value)
+  } else {
+    dims <- length(value)
+  }
```
Bisaloo: If you're using a recent version of R:

```suggestion
  dims <- dim(value) %||% length(value)
```

(you can also redefine the operator in this pkg for compatibility with older R versions)

Artur-man:
Some functions such as `read_h5ad_data_frame` and `read_h5ad_collection` would repeat themselves if we were to define a, e.g. `read_zarr_data_frame/_collection` since they do not directly call `rhdf5` or `Rarr` methods. 

I would instead open a new `read/write_file_helpers.R` to define these cross file format generic methods, and only keep file specific methods like `read_hdf5_numeric_scalar` where they are. 

If you want, I can implement them, otherwise we can explore after you review my changes. Up to you. I am still waiting for some changes in `Huber-group-EMBL/Rarr` but mostly we are ready.

Bisaloo (commented):
FYI, the patterns I highlight here can also be detected via lintr if you use `all_linters()` and then selectively turn off the linters you don't want.

For example: https://github.com/Huber-group-EMBL/Rarr/blob/devel/.lintr

R/write_zarr_helpers.R
Artur-man: oh I will change this directly, anndataR requires 4.5 and later anyways. Thanks Hugo.

R/Rarr_utils.R
Artur-man: done

R/Rarr_utils.R
Artur-man: done

lazappi:
> Some functions such as `read_h5ad_data_frame` and `read_h5ad_collection` would repeat themselves if we were to define a, e.g. `read_zarr_data_frame/_collection` since they do not directly call `rhdf5` or `Rarr` methods.

I think they are slightly different because one will call `read_h5ad_element()` and the other will call `read_zarr_element()`. I'm open to reducing redundancy where it makes sense and is still easy to follow but maybe it's better to leave for a separate PR so it's easier to review the Zarr functionality without other changes.

Artur-man:
Then @lazappi perhaps we can start your review, let me know if you guys want me to change anything (there are lots of changes!). Meanwhile, @Bisaloo will keep working on fixing a few small issues in `Rarr`.

inst/scripts/example_files.py
```diff
@@ -91,3 +102,12 @@
...
 # Write the H5AD file
 adata.write_h5ad("inst/extdata/example.h5ad", compression="gzip")
+
+# Write Zarr 
+adata.write_zarr("inst/extdata/example2.zarr")
+# os.chdir("inst/extdata/")
+# zip = zipfile.ZipFile("example.zarr.zip", "w", zipfile.ZIP_DEFLATED)
+# zip.write("example.zarr")
+# shutil.rmtree("example.zarr")
+# zip.close()
```
lazappi: Maybe this should be uncommented?

inst/scripts/requirements.yml
lazappi, line 1: If we are adding this we can probably remove the versions from the script.

R/AbstractAnnData.R
```diff
@@ -287,6 +287,28 @@ AbstractAnnData <- R6::R6Class(
       )
     },
     #' @description
+    #' Convert to an [`ZarrAnnData`]
```
lazappi:
```suggestion
    #' Convert to a [`ZarrAnnData`]
```

R/AbstractAnnData.R
```diff
@@ -287,6 +287,28 @@ AbstractAnnData <- R6::R6Class(
...
+    #'
+    #' See [as_ZarrAnnData()] for more details on the conversion
+    #'
+    #' @param file See [as_ZarrAnnData()]
+    #' @param compression See [as_ZarrAnnData()]
+    #' @param mode See [as_ZarrAnnData()]
+    #'
+    #' @return An [`ZarrAnnData`] object
```
lazappi:
```suggestion
    #' @return A [`ZarrAnnData`] object
```

R/AbstractAnnData.R
```diff
@@ -302,6 +324,23 @@ AbstractAnnData <- R6::R6Class(
       mode = c("w-", "r", "r+", "a", "w", "x")
     ) {
       write_h5ad(object = self, path, compression = compression, mode = mode)
+    },
+    #' @description
+    #' Write the `AnnData` object to an H5AD file
```
lazappi:
```suggestion
    #' Write the `AnnData` object to a Zarr file
```

R/Rarr_utils.R
```diff
@@ -0,0 +1,75 @@
...
+#'
+#' @param store the location of (zarr) store
+#' @param name name of the group
+#' @param version zarr version
+#' @importFrom utils tail
+#' @examples
+#' store <- tempfile(fileext = ".zarr")
+#' create_zarr(store)
+#' create_zarr_group(store, "gp")
+#' @export
```
lazappi: These are internal functions that shouldn't be exported, use `@noRd` instead

lazappi: Please be more consistent with the formatting of the function documentation, something like:

```suggestion
#' Create Zarr group
#'
#' Create a new group in a Zarr store
#'
#' @param store the location of the Zarr store
#' @param name name of the group to create
#' @param version Zarr version
#'
#' @noRd
#'
#' @examples
#' store <- tempfile(fileext = ".zarr")
#' create_zarr(store)
#' create_zarr_group(store, "gp")
```

R/Rarr_utils.R
```diff
@@ -0,0 +1,75 @@
+#' create_zarr_group
+#'
+#' create zarr groups
+#'
+#' @param store the location of (zarr) store
+#' @param name name of the group
+#' @param version zarr version
+#' @importFrom utils tail
```
lazappi: Imports from common packages can go in `anndataR-package.R`

R/Rarr_utils.R
```diff
@@ -0,0 +1,75 @@
...
+#' @examples
+#' store <- tempfile(fileext = ".zarr")
+#' create_zarr(store)
+#' create_zarr_group(store, "gp")
+#' @export
+#' @return `NULL`
+create_zarr_group <- function(store, name, version = "v2") {
+  split.name <- strsplit(name, split = "/", fixed = TRUE)[[1]]
```
lazappi: Use `snake_case` for variable names

R/Rarr_utils.R
```diff
@@ -0,0 +1,75 @@
...
+  switch(
+    version,
+    v2 = {
+      write(
+        "{\"zarr_format\":2}",
+        file = file.path(store, split.name[1], ".zgroup")
+      )
+    },
+    v3 = {
+      stop("Currently only zarr v2 is supported!")
```
lazappi: Use `cli_abort()` for errors

R/Rarr_utils.R
```diff
@@ -0,0 +1,75 @@
...
+  dir.create(file.path(store, split.name[1]), showWarnings = FALSE)
+  switch(
+    version,
+    v2 = {
+      write(
+        "{\"zarr_format\":2}",
+        file = file.path(store, split.name[1], ".zgroup")
+      )
+    },
+    v3 = {
```
lazappi, line 42: What are your thoughts on being able to support v3?

R/Rarr_utils.R
```diff
@@ -0,0 +1,75 @@
...
+      function(x) paste(split.name[seq_len(x)], collapse = "/"),
+      FUN.VALUE = character(1)
+    )
+    split.name <- rev(tail(split.name, 2))
+    if (!dir.exists(file.path(store, split.name[2]))) {
+      create_zarr_group(store = store, name = split.name[2])
+    }
+  }
+  dir.create(file.path(store, split.name[1]), showWarnings = FALSE)
+  switch(
```
lazappi, line 34: Should this check be in `create_zarr()` (or wherever it will be first triggered)?

R/Rarr_utils.R
```diff
@@ -0,0 +1,75 @@
...
+#'
+#' create zarr store
+#'
+#' @param store the location of zarr store
+#' @examples
+#' store <- tempfile(fileext = ".zarr")
+#' create_zarr(store)
+#' is_zarr_empty(store)
+#' @export
+#' @return returns TRUE if zarr store is not empty
```
lazappi: If the function is named `is_zarr_empty()` it should return `TRUE` when the store is empty. I'm not sure if it's the name or the documentation that is wrong but they don't match.

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+    encoding_list <- read_zarr_encoding(
+      store,
+      name,
+      stop_on_error = stop_on_error
+    )
+    if (is.null(encoding_list)) {
+      if (stop_on_error) {
+        stop("No encoding info found for element '", name, "'")
+      } else {
+        warning("No encoding found for element '", name, "'")
```
lazappi: Use `cli_warn()` for warnings

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+#'
+#' Read the encoding and version of an element in a Zarr store
+#'
+#' @param store A Zarr store instance
+#' @param name Name of the element within the Zarr store
+#'
+#' @return A named list with names type and version
+#'
+#' @noRd
+read_zarr_encoding <- function(store, name, stop_on_error = TRUE) {
```
lazappi: The `stop_on_error` isn't needed here, please make it match the `read_h5ad_encoding()` function

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+  stop_on_error = FALSE,
+  ...
+) {
+  if (is.null(type)) {
+    encoding_list <- read_zarr_encoding(
+      store,
+      name,
+      stop_on_error = stop_on_error
+    )
+    if (is.null(encoding_list)) {
```
lazappi: This check isn't needed because `read_zarr_encoding()` should fail if there is no encoding found

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+    data <- as.vector(data)
+    dim(data) <- length(data)
+  }
+
+  # Reverse {rhdf5} coercion to factors
+  if (is.factor(data) && all(levels(data) %in% c("TRUE", "FALSE"))) {
+    dims <- dim(data)
+    data <- as.logical(data)
+    dim(data) <- dims
+  }
```
lazappi: This probably isn't needed here

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+  version <- match.arg(version)
+
+  # Extract the NestedArray contents as a base R array.
+  data <- read_zarr_array(store, name)
+
+  # If the array is 1D, explicitly add a dimension
+  if (is.null(dim(data))) {
+    data <- as.vector(data)
+    dim(data) <- length(data)
+  }
```
lazappi: This may not be needed depending on what `Rarr::read_zarr_array()` returns

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+    path = file.path(store, name),
+    recursive = FALSE,
+    full.names = FALSE
+  )
+  setNames(
+    lapply(field_names, function(x) {
+      as.vector(Rarr::read_zarr_array(file.path(store, name, x)))
+    }),
+    field_names
+  )
```
lazappi:
```suggestion
  purrr::map(field_names, function(.field) {
    as.vector(read_zarr_array(store, file.path(name, .field)))
  }) |>
    setNames(field_names)
```

lazappi: Maybe this could use `read_zarr_collection()`?

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+  version <- match.arg(version)
+  data <- read_zarr_array(store, name)
+
+  if (is.null(dim(data)) || length(dim(data)) == 1) {
+    data <- as.vector(data)
+    dim(data) <- length(data)
+  }
+
+  # convert "NA" to NA (as in rhdf5:::.h5postProcessDataset)
+  data[data == "NA"] <- NA
```
lazappi, line 291: I don't think `string_array` is nullable but we should check

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+  categories <- read_zarr_array(store, paste0(name, "/categories"))
+
+  # Get codes and convert to 1-based indexing
+  codes <- codes + 1
+
+  if (!length(dim(codes)) == 1) {
+    stop(
+      "There is currently no support for multidimensional categorical arrays"
+    )
+  }
```
lazappi: I don't think this is needed as this can never be written

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+  ordered <- attributes[["ordered"]]
+  if (is.null(ordered) || is.na(ordered)) {
+    warning(
+      "Unable to determine if categorical '",
+      name,
+      "' is ordered, assuming it isn't"
+    )
+
+    ordered <- FALSE
+  }
```
lazappi: I don't think this is needed as the attribute should always be written

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+#' @param store A Zarr store instance
+#' @param name Name of the element within the Zarr store
+#' @param version Encoding version of the element to read
+#' @param include_index Whether or not to include the index as a column
+#'
+#' @details
+#' If `include_index == TRUE` the index stored in the Zarr store is added as a
+#' column to output `data.frame` using the defined index name as the column
+#' name and this is set as an attribute. If `include_index == FALSE` the index
+#' is not provided in the output. In either case row names are not set.
```
lazappi: This should have the same behaviour as `read_h5ad_data_frame()`. The index should always be read and set as the row names. I think the function is correct but the documentation needs updating.

R/write_zarr.R
```diff
@@ -0,0 +1,122 @@
+#' Write Zarr
+#'
+#' Write an Zarr file
```
lazappi:
```suggestion
#' Write a Zarr file
```

R/write_zarr.R
```diff
@@ -0,0 +1,122 @@
...
+#'   * `w` creates a file, truncating any existing ones
+#'   * `w-`/`x` are synonyms creating a file and failing if it already exists
+#' @param ... Additional arguments passed to [as_AnnData()]
+#'
+#' @details
+#'
+#' ## Compression
+#'
+#' Compression is currently not supported for Boolean arrays, they will be
+#' written uncompressed.
```
lazappi: Is this true for Zarr as well?

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,521 @@
...
+        warning("No encoding found for element '", name, "'")
+        return(NULL)
+      }
+    }
+    type <- encoding_list$type
+    version <- encoding_list$version
+  }
+
+  read_fun <- switch(
+    type,
```
lazappi, line 65: Need to add a `read_zarr_null()` function

R/write_zarr_helpers.R
```diff
@@ -0,0 +1,693 @@
...
+  store,
+  name,
+  compression = c("none", "gzip"),
+  stop_on_error = FALSE,
+  ...
+) {
+  compression <- match.arg(compression)
+
+  # Sparse matrices
+  write_fun <-
```
lazappi, line 39: Need to add `write_zarr_null()`

R/write_zarr_helpers.R
```diff
@@ -0,0 +1,693 @@
...
+  store,
+  name,
+  compression,
+  version = "0.2.0"
+) {
+  dims <- dim(value) %||% length(value)
+
+  # replace NA to "NA" (as in rhdf5:::.h5postProcessDataset)
+  # to read as "NA" -> NA later after Rarr:read_zarr_array
+  value[is.na(value)] <- "NA"
```
lazappi, line 380: Don't think we need this

R/ZarrAnnData.R
```diff
@@ -0,0 +1,497 @@
...
+      compression <- match.arg(compression)
+      mode <- match.arg(mode)
+
+      private$.compression <- compression
+
+      private$.close_on_finalize <- is.character(file)
+
+      is_readonly <- FALSE
+
+      if (is.character(file)) {
```
lazappi, line 343: I think for Zarr you only have a path, no handle object? If that is the case some of the logic here can be simplified.

R/ZarrAnnData.R
```diff
@@ -0,0 +1,497 @@
...
+          var <- get_initial_var(var, X, shape)
+          write_empty_zarr(file, obs, var, compression)
+        }
+      }
+
+      # File is supposed to exist by now. Check if it is a valid Zarr file
+      attrs <- Rarr::read_zarr_attributes(file)
+      if (!all(c("encoding-type", "encoding-version") %in% names(attrs))) {
+        cli_abort(c(
+          "File {.file {file}} is not a valid Zarr file."
```
lazappi: Maybe this (and similar messages/documentation) needs to say something like "AnnData Zarr file" because there is no AnnData-specific extension like there is for H5AD.

R/ZarrAnnData.R
```diff
@@ -0,0 +1,497 @@
...
+          }
+        }
+      }
+
+      self
+    },
+
+    # We don't close
+    #' @description Close the Zarr store/file
+    close = function() {},
```
lazappi: This can be removed if it's not needed

R/ZarrAnnData.R
```diff
@@ -0,0 +1,497 @@
...
+    varp = adata$varp,
+    uns = adata$uns,
+    shape = adata$shape(),
+    mode = mode,
+    compression = compression
+  )
+}
+
+# nolint start: object_name_linter
+cleanup_ZarrAnnData <- function(...) {
```
lazappi: I don't think this is used anywhere? It can be removed if so.

tests/testthat/test-h5ad-zarr.R
```diff
@@ -0,0 +1,125 @@
...
+  mat_zarr <- read_zarr_sparse_array(store, "layers/csc_counts", type = "csc")
+  expect_equal(mat_h5ad, mat_zarr)
+
+  mat_h5ad <- read_h5ad_sparse_array(file, "layers/counts", type = "csr")
+  mat_zarr <- read_zarr_sparse_array(store, "layers/counts", type = "csr")
+  expect_equal(mat_h5ad, mat_zarr)
+})
+
+test_that("reading recarrays works", {
+  skip("read_zarr_rec_array is not implemented yet")
```
lazappi: I think this is implemented now?

tests/testthat/test-h5ad-zarr.R
```diff
@@ -0,0 +1,125 @@
...
+})
+
+test_that("reading mappings is same for h5ad and zarr", {
+  skip("for now, example.zarr and example.h5ad are not identical!")
+  mapping_h5ad <- read_h5ad_mapping(file, "uns")
+  mapping_zarr <- read_zarr_mapping(store, "uns")
+  expect_equal(mapping_h5ad, mapping_zarr)
+})
+
+test_that("reading dataframes works", {
```
lazappi:
```suggestion
test_that("reading dataframes is the same for h5ad and zarr", {
```

tests/testthat/test-h5ad-zarr.R
```diff
@@ -0,0 +1,125 @@
...
+  df_h5ad <- read_h5ad_data_frame(file, "obs")
+  df_zarr <- read_zarr_data_frame(store, "obs", include_index = TRUE)
+  expect_equal(df_h5ad, df_zarr)
+})
+
+test_that("reading H5AD as SingleCellExperiment is same for h5ad and zarr", {
+  skip("for now, example.zarr and example.h5ad are not identical!")
```
lazappi: Why are they not identical?

tests/testthat/test-h5ad-zarr.R
```diff
@@ -0,0 +1,125 @@
...
+  expect_equal(mapping_h5ad, mapping_zarr)
+})
+
+test_that("reading dataframes works", {
+  df_h5ad <- read_h5ad_data_frame(file, "obs")
+  df_zarr <- read_zarr_data_frame(store, "obs", include_index = TRUE)
+  expect_equal(df_h5ad, df_zarr)
+})
+
+test_that("reading H5AD as SingleCellExperiment is same for h5ad and zarr", {
```
lazappi: Should probably add a test for Seurat

vignettes/articles/software_design.Rmd
```diff
@@ -50,7 +50,7 @@ The different `AnnData` classes provide a consistent interface but store and acc
   It is primarily designed as an intermediate object when reading/writing H5AD files but can be useful for accessing parts of large files.
 - The `ReticulateAnnData` access data stored in an `AnnData` object in a concurrent Python session.
   This comes with the overhead and complexity of using `r CRANpkg("reticulate")` but is sometimes useful to access functionality that has not yet been implemented in `r Biocpkg("anndataR")`.
-- The planned `ZarrAnnData` will provide an interface to an `AnnData` Zarr store, similar to `HDF5AnnData`.
+- The `ZarrAnnData` provides an interface to a Zarr data (i.e. Zarr store) and, similar to a H5AD file, minimal data is stored in memory until it is requested by the user. It is as an intermediate object when reading/writing Zarr store but can be useful for accessing parts of large files.
```
lazappi:
```suggestion
- The `ZarrAnnData` provides an interface to a Zarr store and, similar to a H5AD file, minimal data is stored in memory until it is requested by the user. It is as an intermediate object when reading/writing Zarr stores but can be useful for accessing parts of large files.
```

lazappi (changes requested):
We will probably need some more rounds of reviews but hopefully I've caught a lot of stuff already.

Additional comments:
- I commented on some style things the first time I saw them but please do the same wherever else it is relevant
- The documentation in `AnnData-usage` needs to be updated to add `ZarrAnnData`
- Round trip tests for Zarr need to be added
- We should probably show Zarr usage somewhere in a main vignette
- Please try to get the CI checks to pass

Artur-man:
> We will probably need some more rounds of reviews but hopefully I've caught a lot of stuff already.
> 
> Additional comments:
> 
> * I commented on some style things the first time I saw them but please do the same wherever else it is relevant
> * The documentation in `AnnData-usage` needs to be updated to add `ZarrAnnData`
> * Round trip tests for Zarr need to be added
> * We should probably show Zarr usage somewhere in a main vignette
> * Please try to get the CI checks to pass

awesome @lazappi I will get to it.

R/read_zarr_helpers.R
Artur-man: This we have discussed with @Bisaloo to how to implement. Currently, rhdf5 forces NAs to "NA" when saved as string arrays, and then return as NA back when returned from the `.h5`.

This doesnt happen in `Rarr`, so we imitate in `anndataR` instead. 

See more here: https://github.com/Huber-group-EMBL/Rarr/issues/86

inst/scripts/requirements.yml
Bisaloo: I learned about this recently, which can be a good lightweight way to document versions without having to create a dedicated environment file: https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

Note that it's not uv-specific but it's also [documented in the official python docs](https://packaging.python.org/en/latest/specifications/inline-script-metadata/#inline-script-metadata).

R/read_zarr_helpers.R
Bisaloo: `Rarr::read_zarr_array()` currently always returns arrays (i.e., `dim()` is not `NULL`). I don't expect this to change as I like the idea of a consistent output class.

README.md
```diff
@@ -54,6 +54,9 @@ the task you want to perform.
   `BiocManager::install("SingleCellExperiment")`
 - To convert to/from `Seurat` objects, install [SeuratObject](https://cran.r-project.org/package=SeuratObject):  
   `install.packages("SeuratObject")`
+- To read/write \*.zarr files, you need to install
+  [zarr](https://www.bioconductor.org/packages/release/bioc/html/Rarr.html):
```
Bisaloo:
```suggestion
  [Rarr](https://www.bioconductor.org/packages/release/bioc/html/Rarr.html):  
```

R/Rarr_utils.R
Artur-man: @Bisaloo is working on it, i think he wants to do this until march-april

README.md
Artur-man: fixed

vignettes/articles/software_design.Rmd
Artur-man: fixed

tests/testthat/test-h5ad-zarr.R
Artur-man: some readers are not implemented yet, but will check again.
https://github.com/scverse/anndataR/pull/190#issuecomment-3619051125

tests/testthat/test-h5ad-zarr.R
Artur-man: fixed

tests/testthat/test-h5ad-zarr.R
Artur-man: Hmm, ok the skip here is misleading, since `Rarr` currently doesn't read arbitrary structured data types, but my current implementation under `read_zarr_rec_array` only works for the example data. So it should stay like this until the following is fixed: 
https://github.com/Huber-group-EMBL/Rarr/issues/81

R/AbstractAnnData.R
Artur-man: fixed

R/AbstractAnnData.R
Artur-man: fixed

R/Rarr_utils.R
Artur-man: Hmm, good idea but I think it doesn't matter where it is put. Either in `create_zarr` or `create_zarr_group` this check should be performed. Since root zarr is a zarr group too, we call `create_zarr_group` with `create_zarr`. 

However, I like how you think. This functions are auxiliary and will be deprecated once @Bisaloo implements it in `Rarr`.

R/read_zarr_helpers.R
Artur-man: lets leave this for now, because we are still waiting @Bisaloo to implement readers for structured data. 
https://github.com/Huber-group-EMBL/Rarr/issues/81

Artur-man:
Errors I am getting at lint and pkgdown GHAs look unrelated to the package: 

```
Error in dyn.load(file, DLLpath = DLLpath, ...) : 
    unable to load shared object '/home/runner/work/_temp/Library/Rhdf5lib/libs/Rhdf5lib.so':
    libsz.so.2: cannot open shared object file: No such file or directory
```

Artur-man:
@Bisaloo also the tests that are passing for macos, ubuntu and windows are failing for macos (intel) and ubuntu-bioc-release. I understand the error in release (for obvious reasons), but macos-15 I don't know. Any ideas ? 

```
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
   5.         └─base::array(metadata$fill_value, dim = lengths(index))
  ── Error ('test-Zarr-read.R:87:3'): reading numeric scalars works ──────────────
  Error in `array(metadata$fill_value, dim = lengths(index))`: 'dim' cannot be of length 0
  Backtrace:
      ▆
   1. └─anndataR:::read_zarr_numeric_scalar(store, "uns/IntScalar") at test-Zarr-read.R:87:3
   2.   └─anndataR:::read_zarr_array(store, name)
   3.     └─Rarr::read_zarr_array(file.path(store, name))
   4.       └─Rarr:::read_data(...)
   5.         └─base::array(metadata$fill_value, dim = lengths(index))
  
  [ FAIL 4 | WARN 447 | SKIP 12 | PASS 1842 ]
```

https://github.com/scverse/anndataR/actions/runs/20411185095/job/58648259569?pr=190

Bisaloo:
Yes, it looks that on this runner, Rarr 1.11.9 is installed, when support for reading scalars was added in Rarr 1.11.12. I think it's related to BioC devel not having more recent binaries for macOS x86 yet.

Maybe try setting:

```yaml
Imports:
    Rarr (>= 1.11.12)
```

In `DESCRIPTION` to force installation from source?

This will not work for the runner using BioC release but it's already failing anyways.

Artur-man:
I see. Then I will assume all checks are passing for now. thanks man

R/Rarr_utils.R
Artur-man: fixed

R/Rarr_utils.R
Artur-man: ok, I think it is fixed.

R/Rarr_utils.R
Artur-man: fixed

R/Rarr_utils.R
Artur-man: fixed

R/read_zarr_helpers.R
Artur-man: fixed

R/read_zarr_helpers.R
Artur-man: fixed

R/read_zarr_helpers.R
Artur-man: fixed

R/read_zarr_helpers.R
Artur-man: will leave until https://github.com/Huber-group-EMBL/Rarr/issues/81 is fixed.

R/read_zarr_helpers.R
Artur-man: fixed

R/read_zarr_helpers.R
Artur-man: fixed

R/read_zarr_helpers.R
Artur-man: fixed

R/write_zarr.R
Artur-man: fixed

R/write_zarr_helpers.R
Artur-man: nope, unfortunately the behaviour is not the same as in rhdf5. @Bisaloo wanna comment ? 

```
> store <- tempfile(fileext = ".zarr")
> create_zarr(store = store)
> string <- LETTERS[1:5]
> string[5] <- NA
> write_zarr_element(string, store, "string_array")
NULL
Warning message:
Could not write element 'string_array' of type 'character': missing value where TRUE/FALSE needed 
```

R/read_zarr_helpers.R
Artur-man: fixed, @lazappi are there tests for this for hdf5/h5ad ?

R/write_zarr_helpers.R
Artur-man: need to check with @Bisaloo for this.

Artur-man:
Current issues: 

- [x] https://github.com/Huber-group-EMBL/Rarr/pull/82
- [x] https://github.com/Huber-group-EMBL/Rarr/issues/86
- [x] https://github.com/Huber-group-EMBL/Rarr/issues/81
- [x] implement `write_zarr_null`.
- [x] update `ZarrAnnData$initialize` since Zarr store doesn't have a handler.
- [x] The documentation in AnnData-usage needs to be updated to add ZarrAnnData
- [x] We should probably show Zarr usage somewhere in a main vignette
- [x] https://github.com/Huber-group-EMBL/Rarr/pull/104
- [x] Support for Unicode rec-arrays https://github.com/Huber-group-EMBL/Rarr/issues/81 
- [x] fix roundtrip tests for zarr
- [x] remove `eval=FALSE` from vignette chunks after fixes
- [x] Get the CI checks to pass

R/ZarrAnnData.R
Artur-man: I see makes sense. Fixed

tests/testthat/test-h5ad-zarr.R
Artur-man: fixed, but skipping for now until identical.

inst/scripts/example_files.py
Artur-man: done

R/ZarrAnnData.R
Artur-man: fixed

R/ZarrAnnData.R
Artur-man: fixed

R/ZarrAnnData.R
Artur-man: ok, I added some additional checks, thus now if `file` is not a character then it will give an error. But let me know I have missed something here. 

I also check if the file is a valid zarr folder, which is different than the folder being a valid anndata-zarr.

R/write_zarr_helpers.R
Artur-man: the read and write of null encoding is strictly based on the `.zattrs`. I have tried reading the null array in a zarr v2 implementation and that gives an error. Thus, I will just write an empty logical array here with null encoding in the attr. I will leave a `TODO:` for future correction though.

tests/testthat/test-h5ad-zarr.R
Artur-man: just letting you know, exact equality here is hard to achieve ... this is because of how structed data types (rec arrays are read between h5ad and zarr, and even between h5py and rhdf5):

see here
Huber-group-EMBL/rhdf5#192

tests/testthat/test-h5ad-zarr.R
```diff
@@ -0,0 +1,169 @@
...
+  array_h5ad <- read_h5ad_string_array(file, "uns/String")
+  array_zarr <- read_zarr_string_array(store, "uns/String")
+  expect_equal(array_h5ad, array_zarr)
+
+  array_h5ad <- read_h5ad_string_array(file, "uns/String2D")
+  array_zarr <- read_zarr_string_array(store, "uns/String2D")
+  expect_equal(array_h5ad, array_zarr)
+})
+
+test_that("reading mappings is same for h5ad and zarr", {
```
Artur-man: There is a difference between rhdf5, Rarr and h5py on how rec arrays and structured data types are read. 

To me, h5py and Rarr reads it correctly, but rhdf5 does not:
Huber-group-EMBL/rhdf5#192

Artur-man:
We are really close, tests are passing on my local but we have to wait until changes in `Rarr` are in BioC-devel. 
Let me give a break I will review the entire PR. Please also let me know if you catch stuff, otherwise I will get back to this to finalize. 

Thanks @Bisaloo again for answering all issues of Rarr!

tests/testthat/test-h5ad-zarr.R
lazappi: Maybe we need an issue for this here as well? From our point of view it doesn't matter so much what is returned by **{rhdf5}** as long as what we return from the `AnnData` matches Python.

Going the other way (writing a `recarray`) is more difficult because we don't know when to use that format. I don't think we test this. Also, I don't think the `AnnData` disk spec really covers alternative arrays like `recarray`.

Artur-man: https://github.com/scverse/anndataR/issues/409

tests/testthat/test-h5ad-zarr.R
Artur-man: should be done now

tests/testthat/test-h5ad-zarr.R
Artur-man: mostly done, but also being addressed in https://github.com/scverse/anndataR/issues/409

tests/testthat/test-h5ad-zarr.R
Artur-man: done

Bisaloo:
As discussed in the bioconductor zulip, you're not likely to get up-to-date macOS x86 binaries for Rarr. So you'll need to either force installing from source, or add the bioc r-universe as an additional repository for this test to pass.

Alternatively, setting 1.11.24 as minimum required version for Rarr might do the trick to force installing from source when a recent enough binary isn't available.

inst/scripts/requirements.yml
Artur-man: Hmm, I actually like this: 
https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

Let me know what you think @lazappi, I can embed it and drop yml ... but I do not know if this would make development harder for people used to running conda env through yml.

R/Rarr_utils.R
Artur-man: I would keep the check for zarr format here (until we can write zarr groups/arrays directly). Since zarr store is already a zarr root group and this check should be done for groups too. 

That being said, since we do not define higher-level classes for zarr stores (and instead open/create/modify groups and arrays directly via path), we cannot check if, for example, if the original zarr store is v3 or while the group you are trying to create is v2, which should normally throw an error. 

However, all these requires major changes to Rarr and its inner workings.

R/Rarr_utils.R
Artur-man: We can read zarr v3 now ... but not writing, perhaps this is enough for pushing zarr support to anndataR release.

inst/scripts/requirements.yml
lazappi: Yeah, I think I prefer this to having a separate file as well

R/Rarr_utils.R
lazappi: Ok. I think the way this is set up is reasonable. The argument makes it clear we know there are versions and the message should be obvious that only v2 is currently supported. I would adjust the messages a bit but I will suggest something later.

lazappi:
> As discussed in the bioconductor zulip, you're not likely to get up-to-date macOS x86 binaries for Rarr. So you'll need to either force installing from source, or add the bioc r-universe as an additional repository for this test to pass.
> 
> Alternatively, setting 1.11.24 as minimum required version for Rarr might do the trick to force installing from source when a recent enough binary isn't available.

Could you please explain what the issue with this is? Is it something on the Bioconductor side so do we need to look into it?

I'm guessing the current release check is also expected to fail because we need a later version of **{Rarr}**?

Bisaloo:
> Could you please explain what the issue with this is? Is it something on the Bioconductor side so do we need to look into it?

BioC macOS x86 builder is no longer working. They're planning on pulling the binaries from r-universe, as they are doing with windows binaries, but it's still a work in progress.
In the meantime, the latest available version for macOS x86 binaries in BioC devel is quite outdated.
It's something that needs to be (and will be) resolved by Bioconductor.

But you can still work around the issue if you need to. By requiring installation from source (this should happen automatically if you request a version more recent than what the binaries provide). Or by getting the binaries from r-universe yourself.

> I'm guessing the current release check is also expected to fail because we need a later version of {Rarr}?

Yes, exactly

inst/extdata/example_v2.zarr.zip
rcannood, line 1: Can we Rbuildignore the example_v2.zarr.zip and example_v3.zarr.zip files?

inst/scripts/requirements.yml
```diff
@@ -0,0 +1,14 @@
...
+dependencies:
+  - python=3.13.5
+  - pip
+  - pip:
+      - anndata==0.12.10
+      - scanpy==1.12.0
+      - scipy==1.17.1
+      - zarr==3.1.5
+      - igraph==1.0.0
+      - leidenalg==0.11.0
```
rcannood:
```suggestion
      - leidenalg==0.11.0

```

Add missing newline

R/AbstractAnnData.R
```diff
@@ -313,6 +335,37 @@ AbstractAnnData <- R6::R6Class(
...
+        "zstd",
+        "lzma",
+        "bz2",
+        "zlib",
+        "lz4"
+      ),
+      mode = c("w-", "r", "r+", "a", "w", "x")
+    ) {
+      write_zarr(
+        object = self,
```
rcannood: fix linting issues

R/AbstractAnnData.R
```diff
@@ -313,6 +335,37 @@ AbstractAnnData <- R6::R6Class(
...
+    #' See [write_zarr()] for details
+    #'
+    #' @param path See [write_zarr()]
+    #' @param compression See [write_zarr()]
+    #' @param mode See [write_zarr()]
+    #'
+    #' @return `path` invisibly
+    write_zarr = function(
+      path,
+      compression = c(
```
rcannood, line 360: the compression arguments of `write_zarr` is different from `as_ZarrAnnData`. which one is correct?

R/read_h5ad_helpers.R
```diff
@@ -179,7 +179,6 @@ read_h5ad_element_keys <- function(
 #' @noRd
 read_h5ad_null <- function(file, name, version = "0.1.0") {
   version <- match.arg(version)
-
   NULL
```
rcannood, line 183: undo this unnecessary change to avoid merge conflicts

tests/testthat/test-roundtrip-empty-zarr.R
rcannood, line 1: merge `test-roundtrip-*-zarr.R` into `test-roundtrip-*.R`

vignettes/articles/software_design.Rmd
```diff
@@ -23,34 +23,30 @@ knitr::opts_chunk$set(
 * [_anndata_](https://anndata.readthedocs.io/en/stable/) ([scverse/anndata](https://github.com/scverse/anndata)): The Python `anndata` package and on-disk specification
 * `r Biocpkg("zellkonverter")` ([theislab/zellkonverter](https://github.com/theislab/zellkonverter)): Convert `AnnData` files to/from `SingleCellExperiment` objects
 * `r Githubpkg("mtmorgan/h5ad")`: Read/write `*.h5ad` files natively using `r Biocpkg("rhdf5")`
-* `r CRANpkg("anndataR")` ([dynverse/anndata](https://github.com/dynverse/anndata)): An R implementation of the `AnnData` data structures, uses `r CRANpkg("reticulate")` to read/write `*.h5ad` files
+* `r Biocpkg("Rarr")`: Read/write `*.zarr` stores natively
```
rcannood: Rarr is a dependency, but is not a package related to reading or writing anndata files

rcannood (commented):
First round of comments prior to the hackathon!

Please merge devel again

inst/extdata/example_v2.zarr.zip
Artur-man: what about example.h5ad ?

R/AbstractAnnData.R
Artur-man: are you talking about the possible values of `compression`. I think they are same already? 

```
as_ZarrAnnData <- function(
  # nolint end: object_name_linter
  adata,
  file,
  compression = c(
    "none",
    "gzip",
    "blosc",
    "zstd",
    "lzma",
    "bz2",
    "zlib",
    "lz4"
  ),
  mode = c("w-", "r", "r+", "a", "w", "x")
```

```
write_zarr <- function(
  object,
  path,
  compression = c(
    "none",
    "gzip",
    "blosc",
    "zstd",
    "lzma",
    "bz2",
    "zlib",
    "lz4"
  ),
  mode = c("w-", "r", "r+", "a", "w", "x"),
  ...

File: vignettes/articles/software_design.Rmd (hunk 47 above)
Comment (Artur-man) [reply vignettes/articles/software_design.Rmd]: Hmm, ok you are right, so delete then ? I cannot think of an implementation to be inspired from based on `Rarr`.

File: R/read_h5ad_helpers.R (hunk 46 above)
Comment (Artur-man) [reply R/read_h5ad_helpers.R:183]: done

File: inst/scripts/requirements.yml (hunk 43 above)
Comment (Artur-man) [reply inst/scripts/requirements.yml]: done

File: inst/extdata/example_v2.zarr.zip
Comment (lazappi) [reply inst/extdata/example_v2.zarr.zip:1]: We use these in the examples/tests so they need to be in the package

File: inst/scripts/example_files.py [hunk 48]
```diff
@@ -91,3 +108,25 @@
...
+        for root, dirs, files in os.walk(zarr_dir):
+            for file in files:
+                full_path = os.path.join(root, file)
+                arcname = os.path.relpath(
+                    full_path,
+                    start=os.path.dirname(zarr_dir)
+                )
+                z.write(full_path, arcname)
+
+    shutil.rmtree(zarr_dir)
```
lazappi, line 157: Can this be simplified?

R/AbstractAnnData.R
lazappi: This should be fixed now

R/Rarr_utils.R
```diff
@@ -0,0 +1,111 @@
+#' create_zarr_group
+#'
+#' Create a zarr group
+#'
+#' @param store the location of (zarr) store
+#' @param name name of the group
+#' @param version zarr version
```
lazappi: Parameter descriptions should use sentence case e.g. `@param param Param does this`

R/Rarr_utils.R
```diff
@@ -0,0 +1,111 @@
...
+#'
+#' @return `NULL`
+#'
+#' @noRd
+#'
+#' @examples
+#' store <- tempfile(fileext = ".zarr")
+#' create_zarr(store)
+#' create_zarr_group(store, "gp")
```
lazappi: Don't need examples on unexported functions

R/Rarr_utils.R
```diff
@@ -0,0 +1,111 @@
...
+      seq_along(split_name),
+      function(x) paste(split_name[seq_len(x)], collapse = "/"),
+      FUN.VALUE = character(1)
+    )
+    split_name <- rev(tail(split_name, 2))
+    if (!dir.exists(file.path(store, split_name[2]))) {
+      create_zarr_group(store = store, name = split_name[2])
+    }
+  }
+  dir.create(file.path(store, split_name[1]), showWarnings = FALSE)
```
lazappi, line 33: Add comments explaining this

R/Rarr_utils.R
lazappi: I think we should delay adding v3 write support to a future PR (when all the required functionality is available)

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,481 @@
...
+    paste0(name, "/data")
+  )))
+  indices <- as.vector(Rarr::read_zarr_array(file.path(
+    store,
+    paste0(name, "/indices")
+  )))
+  indptr <- as.vector(Rarr::read_zarr_array(file.path(
+    store,
+    paste0(name, "/indptr")
+  )))
```
lazappi: Fix identation

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,481 @@
...
+  } else if (type == "csr_matrix") {
+    mtx <- Matrix::sparseMatrix(
+      j = indices,
+      p = indptr,
+      x = data,
+      dims = shape,
+      repr = "R",
+      index1 = FALSE
+    )
+  }
```
lazappi: Change to how this is done for H5AD. Could move to a function.

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,481 @@
...
+  version <- match.arg(version)
+
+  items <- list.dirs(
+    path = file.path(store, name),
+    recursive = FALSE,
+    full.names = FALSE
+  )
+
+  # Omit Zarr metadata files from the list of columns.
+  items <- items[!items %in% c(".zgroup", ".zattrs", ".zarray")]
```
lazappi: Maybe we should store the metadata file vector somewhere and reuse it?

R/read_zarr_helpers.R
```diff
@@ -0,0 +1,481 @@
...
+
+  # h5ad like ordering, see H5_ITER_INC
+  items <- H5_ITER_INC_ORDERING(items)
```
lazappi: Why is this needed?

R/write_zarr_helpers.R
lazappi: We should check this matches whatever Python does

R/write_zarr_helpers.R
lazappi: The main thing is not that our Zarr and H5AD backends are the same but that they are both consistent with their matching Python implementations

R/write_zarr_helpers.R
```diff
@@ -0,0 +1,728 @@
...
+    "none" = NULL,
+    "zstd" = Rarr::use_zstd(),
+    "blosc" = Rarr::use_blosc(),
+    "gzip" = Rarr::use_gzip(),
+    "lzma" = Rarr::use_lzma(),
+    "bz2" = Rarr::use_bz2(),
+    "zlib" = Rarr::use_zlib(),
+    "lz4" = Rarr::use_lz4()
+  )
+}
```
lazappi: Document this, use a better variable name

R/ZarrAnnData.R
```diff
@@ -0,0 +1,501 @@
...
+#' @description
+#' Implementation of a Zarr-backed `AnnData` object. This class provides an
+#' interface to a Zarr file and minimal data is stored in memory until it is
+#' requested by the user. It is primarily designed as an intermediate object
+#' when reading/writing Zarr files but can be useful for accessing parts of
+#' large files.
+#'
+#' See [AnnData-usage] for details on creating and using `AnnData` objects.
+#'
+#' @return An `ZarrAnnData` object
```
lazappi: Use "A `ZarrAnnData`" not "An `ZarrAnnData`"

```suggestion
#' @return A `ZarrAnnData` object
```

R/ZarrAnnData.R
```diff
@@ -0,0 +1,501 @@
...
+  cloneable = FALSE,
+  private = list(
+    .zarrobj = NULL,
+    .compression = NULL,
+
+    .check_file_valid = function() {
+      if (!zarr_path_exists(private$.zarrobj, "/")) {
+        cli_abort(
+          paste(
+            "The Zarr file does not exist, or not a zarr file/store!"
```
lazappi:
```suggestion
            "The Zarr path does not exist or is not a valid Zarr store"
```

tests/testthat/test-h5ad-zarr.R
lazappi, line 1: Restructure this to use a similar approach as the roundtrip tests

tests/testthat/test-Zarrv3-read.R
lazappi, line 1: Merge this into `test-Zarr-read.R`

vignettes/articles/software_design.Rmd
```diff
@@ -30,15 +30,11 @@ Ideally, this package will be a complete replacement for all of these packages,
 # Core features
 
 * An `r CRANpkg("R6")` `AnnData` class to work with objects in R
-* In-memory (`InMemoryAnnData`), HDF5-backed (`HDF5AnnData`) and Python-backed (`ReticulateAnnData`) back ends with a consistent interface
-* Read/write `.h5ad` files natively
+* In-memory (`InMemoryAnnData`), HDF5-backed (`HDF5AnnData`), Zarr-backed (`ZarrAnnData`), Python-backed (`ReticulateAnnData`) back ends with a consistent interface
```
lazappi:
```suggestion
* In-memory (`InMemoryAnnData`), HDF5-backed (`HDF5AnnData`), Zarr-backed (`ZarrAnnData`) and Python-backed (`ReticulateAnnData`) back ends with a consistent interface
```

vignettes/anndataR.Rmd
```diff
@@ -36,9 +36,9 @@ This package builds on our experience developing and using other interoperabilit
 
 Existing packages provide similar functionality to `r Biocpkg("anndataR")` but there are some important differences:
 
-- `r Biocpkg("zellkonverter")` provides conversion of `SingleCellExperiment` objects to/from `AnnData` and reading/writing of `.h5ad` files.
-  This is facilitated via `r CRANpkg("reticulate")` using `r Biocpkg("basilisk")` to manage Python environments (native reading of `.h5ad` files is also possible).
-  In contrast, `r Biocpkg("anndataR")` provides a native R H5AD interface, removing the need for Python dependencies.
+- `r Biocpkg("zellkonverter")` provides conversion of `SingleCellExperiment` objects to/from `AnnData` and reading/writing of `.h5ad` files and `.zarr` stores.
+  This is facilitated via `r CRANpkg("reticulate")` using `r Biocpkg("basilisk")` to manage Python environments (native reading of `.h5ad` and `.zarr` is also possible).
+  In contrast, `r Biocpkg("anndataR")` provides a native R H5AD and Zarr interface, removing the need for Python dependencies.
```
lazappi:
```suggestion
- `r Biocpkg("zellkonverter")` provides conversion of `SingleCellExperiment` objects to/from `AnnData` and reading/writing of `.h5ad`.
  This is facilitated via `r CRANpkg("reticulate")` using `r Biocpkg("basilisk")` to manage Python environments (native reading of `.h5ad` is also possible).
  In contrast, `r Biocpkg("anndataR")` provides a native R H5AD and Zarr interface, removing the need for Python dependencies.
```

vignettes/anndataR.Rmd
```diff
@@ -97,6 +97,31 @@ There is also a HDF5-backed `AnnData` object:
...
+# as SingleCellExperiment
+sce <- read_zarr(zarr_path, as = "SingleCellExperiment")
+
+# as SingleCellExperiment
+obj <- read_zarr(zarr_path, as = "Seurat")
+
+# as Zarr-backed
+adata <- read_zarr(zarr_path, as = "ZarrAnnData")
+```
+
```
lazappi, line 122: These examples should be evaluated

vignettes/anndataR.Rmd
```diff
@@ -187,6 +212,19 @@ tmpfile <- tempfile(fileext = ".h5ad")
...
+tmpfile <- tempfile(fileext = ".zarr")
+adata$write_zarr(tmpfile) # Alternatively, write_zarr(adata, tmpfile)
+
+tmpfile <- tempfile(fileext = ".zarr")
+write_zarr(sce, tmpfile)
+
+tmpfile <- tempfile(fileext = ".zarr")
+write_zarr(obj, tmpfile)
+```
+
```
lazappi, line 225: Should be evaluated

vignettes/usage_seurat.Rmd
```diff
@@ -58,6 +58,27 @@ seurat_obj <- adata$as_Seurat()
...
+seurat_obj_zarr
+```
+
+or 
+
+```{r convert-seurat-zarr, eval=FALSE}
+adata <- read_zarr(zarr_path)
+seurat_obj_zarr <- adata$as_Seurat()
+seurat_obj_zarr
+```
```
lazappi: Evaluate examples

vignettes/usage_singlecellexperiment.Rmd
```diff
@@ -61,6 +61,27 @@ sce <- adata$as_SingleCellExperiment()
...
+sce_zarr
+```
+
+or 
+
+```{r read-zarr, eval=FALSE}
+adata <- read_zarr(zarr_path)
+sce_zarr <- adata$as_SingleCellExperiment()
+sce_zarr
+```
```
lazappi: Evaluate examples

lazappi (commented):
A lot of these are TODOs for me 😸

inst/scripts/example_files.py
Bisaloo: FYI, I'm planning to write a separate, more full-featured, package to handle Zarr stores after the next Bioconductor release and it'll handle zip stores out of the box, which could make you life a bit easier here.

R/write_zarr_helpers.R
lazappi: I think this behaviour might actually be incorrect for H5AD, I've created an issue to follow up https://github.com/scverse/anndataR/issues/444

Artur-man:
@lazappi I think I did not check before finishing the hackathon. Do you set the default zarr version for write methods now ? or do you write only zarr v2 or v3 ? 

Again here is the parameter from `anndata` to set the default behaviour:
https://github.com/keller-mark/anndataR/blob/528eff0024f73b39c5147d5eb15e7ad8b9b7bade/inst/scripts/example_files.py#L142

would be nice to do that too if we are not rushing bioc release.

Artur-man:
Ok I took another look at it. The roundtrip tests and write tests do not have zarr v3 tests but thats fine for now I think. What is more important is that we can read both versions. Since this is gonna stay as the devel branch for another 6 months I can throw another PR later. Lets not overdo it. 

**UPDATE**: I wrote a version of `test-Zarr-write.R` loops over both v2 and v3 ... looks like its working.

lazappi:
I thought there was something that was missing from **{Rarr}** that was blocking writing v3 but I forgot to open an issue and come back to it. If you think the extra tests are passing do you want to push them here?

lazappi:
@keller-mark @Artur-man @Bisaloo 

Thanks for your contributions to this PR so far! To properly recognise them feel free to add yourselves as contributors to the `DESCRIPTION`. Or if it's easier comment with your details and I will add you.

DESCRIPTION
```diff
@@ -18,6 +18,8 @@ Authors@R: c(
            comment = c(ORCID = "0000-0001-5949-2191", github = "bschilder")),
     person("Chananchida", "Sang-aram", role = "ctb",
            comment = c(ORCID = "0000-0002-0922-0822", github = "csangara")),
+    person("Artür", "Manukyan", role = "ctb",
+           comment = c(ORCID = "0000-0002-0441-9517", github = "Artur-man")),
```
Bisaloo, line 22:
```suggestion
           comment = c(ORCID = "0000-0002-0441-9517", github = "Artur-man")),
    person("Hugo", "Gruson", role = "ctb",
           comment = c(ORCID = "0000-0002-4094-1476", github = "Bisaloo")),
```

Thanks for the offer!

rcannood (approved):
LGTM!

rcannood:
Looks good, thank you all!

The CI is failing, but that's because of the Rarr minimal version requirement. When Bioconductor 3.23 is released, this issue should be fixed.

Artur-man:
We did it!!!  1.5 years ... thanks for everyone for the support @keller-mark @Bisaloo @rcannood @lazappi @LouiseDck

ilan-gold:
Congrats all!

PR #193, opened 2024-11-07
experiment with new dummy anndata

rcannood:
@LouiseDck : To begin with, what would you think about the following as roundtrip tests? This would require the least amount of refactoring in comparison to the current unit tests, and would allow us to test our reading and writing capabilities.

WDYT?


tests/testthat/test-roundtrip-obsvar.R
```diff
@@ -1,102 +1,72 @@
...
     )
+
     expect_equal(
-      ad_new$var[[name]],
-      data$var[[name]],
+      adata_py2$var[[name]],
+      adata_py$var[[name]],
       ignore_attr = TRUE,
       tolerance = 1e-6
     )
```
LouiseDck, line 72: What is the goals of these tests? 
Correct me if I'm wrong, but this just checks whether the cran anndata correctly reads the file, and I'm not sure why that should be tested?

tests/testthat/test-roundtrip-obsvar.R
```diff
@@ -1,102 +1,72 @@
...
-    ad_new <- read_h5ad(filename, to = "HDF5AnnData")
+    adata_r <- read_h5ad(filename, to = "HDF5AnnData")
 
     # expect slots are unchanged
     expect_equal(
-      ad_new$obs[[name]],
-      data$obs[[name]],
+      adata_r$obs[[name]],
+      adata_py$obs[[name]],
       ignore_attr = TRUE,
```
LouiseDck, line 42: Why do you want to ignore attributes (I don't mean we shouldn't, just asking)

tests/testthat/test-roundtrip-obsvar.R
```diff
@@ -1,102 +1,72 @@
...
       tolerance = 1e-6
```
LouiseDck, line 43: I think we probably should detect small data changes and signal them to the user instead of ignoring them, if there is a difference between the precision of certain datatypes 
-- this is something that can happen at a later state, but we should keep it in mind?

LouiseDck (commented):
One thing that was tested originally, was writing & reading and expecting the same results. I don't think this is tested with the current tests, and I think it is still a nice feature to test consistency of the writing and reading functions?

tests/testthat/test-roundtrip-obsvar.R
rcannood: Cran anndata is not being used here -- it tests whether the file written by anndataR is still the same as the original file created by python anndata.

However, as is, the data is being transferred over reticulate, which we want to avoid. I'll make changes to fix accordingly!

LouiseDck: Right, I missed that `ad <- reticulate::import("anndata")` refers to the Python anndata, sorry. But I agree about transferring over reticulate!

rcannood: Fixed!

tests/testthat/test-roundtrip-obsvar.R
rcannood: Sometimes attributes are added when you don't expect them. At any rate, I don't compare the R objects to Python anymore, so this code was removed :relaxed:

tests/testthat/test-roundtrip-obsvar.R
rcannood: I don't think inherent issues with floating point arithmetic are relevant to report on, but rather it's something users need to be generally aware of when using FPs.

rcannood:
@lazappi do you have any suggestions for this PR? The purpose of the PR is to have a roundtrip test for anndataR which doesn't use reticulate's py_to_r or r_to_py (but does use reticulate to execute python code)

lazappi:
I think the structure of it looks ok (some of the variables could have better names though 😸). From what I can see, this tests writing a file with Python and reading it into R, and writing and file with R and reading it in Python. I think that addresses @LouiseDck's comment? I guess you could make the object more complex to test more things but that could also be added later. Was there something more specific you wanted me to look at?

rcannood:
That sums it up, indeed!

Ideally, I'd directly check whether what R reads in is the same as what python had written, because we don't check whether the R matrix we get out of reading X (for instance) has the right values, only that we are able to reproduce the same data after we write it back out again.

However, this approach is one way to be able to get reticulate out of the equation to make sure that the errors we observe are not due to reticulate conversions.

Other data structures (varm, etc) will be covered in a separate PR.

tests/testthat/test-roundtrip-obsvar.R
LouiseDck: Sure, if the tests in #170 catch the issues wrt issues like **{anndataR}** storing as float32 and **{anndata}** storing as float64

PR #196, opened 2024-11-12
return NULL if path does not exist

rcannood:
Without this, we get an error when reading an empty h5ad.

## Example

```python
import anndata as ad

adata = ad.AnnData(
  shape=(10, 20)
)

adata.write_h5ad("empty.h5ad")
```

```r
devtools::load_all()
adata <- read_h5ad("empty.h5ad")
```

    Error in `[[.H5File`(file, name) : 
      An object with name X does not exist in this group

## Testing

This situation will be covered by #193 by removing `X = ...` from the dummy_anndata generate_dataset call

### DIFF
--- a/R/read_h5ad_helpers.R
+++ b/R/read_h5ad_helpers.R
@@ -49,7 +49,7 @@ read_h5ad_encoding <- function(file, name) {
 #'
 #' @noRd
 read_h5ad_element <- function(file, name, type = NULL, version = NULL, stop_on_error = FALSE, ...) {
-  if (!hdf5_path_exists(name)) {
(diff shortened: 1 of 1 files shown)


lazappi:
How much have you tested this? I wonder if returning `NULL` could break some things, like do we use `X` to get `n_vars` and `n_obs`?

rcannood:
Since X might not always be in the hdf5 file (as in, the group does not even exist), we should use it to derive values like n_obs and n_vars.

Just noticed that I made a mistake when committing my code, hopefully fixed now.

lazappi (approved):
I checked and it's `obs`/`var` that have to exist so I think this should be fine

PR #197, opened 2024-11-12
Fix AnnData initialisation when var is NULL

### LINKED ISSUE #184: Initialization of AnnData object fails from named matrix
Hi,
I am running into an error while trying to construct an object from Seurat. 
(Trying to work around anndataR currently not working correctly with Seurat v5 objects)

```
matrix <- t(srat[["RNA"]]$counts)
adata <- AnnData(
  X = matrix,

)

```

Minimal example:

```
library(anndataR)


mtx <- matrix(rnorm(100), nrow = 10)
rownames(mtx) <- paste0("cell", 1:10)
colnames(mtx) <- paste0("gene", 1:10)

adata <- AnnData(
  X = mtx
)
>>> Error in names(x) <- value : 
  'names' attribute [10] must be the same length as the vector [0]
Calls: AnnData ... <Anonymous> -> initialize -> get_initial_var -> colnames<-
```

How do I handle this properly?

rcannood:
Fixes #184

This PR fixes an issue when creating a new AnnData when X is defined but var is undefined.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -49,7 +49,7 @@ get_initial_var <- function(var, X, shape) {
   if (is.null(var)) {
     var <- data.frame(matrix(NA, nrow = shape[[2]], ncol = 0))
     if (!is.null(X)) {
-      colnames(var) <- colnames(X)
+      rownames(var) <- colnames(X)
     }
   }
   var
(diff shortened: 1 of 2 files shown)


rcannood:
Merging the fix without a review since the proposed changes are quite evident

PR #211, opened 2024-12-12
add list of known issues to the design rmd

rcannood:
* Add a list of known issues to the design Rmd

This is what it looks like:

Screenshot from 2024-12-12 22-51-54

I'm starting to think it may be better to split the design Rmd into: design, feature tracking, known issues. WDYT?


LouiseDck (approved):
Looks awesome, great to have this documented on the website too!

PR #212, opened 2024-12-13
Conversion from & to SingleCellExperiment

LouiseDck:
Provides modular conversion to & from SingleCellExperiment.

Adds vignette explaining the mapping in detail.

Aims to follow some of the mapping of the Seurat conversion.


R/SingleCellExperiment.R
```diff
@@ -27,67 +74,208 @@
...
+# nolint start: cyclocomp_linter
+to_SingleCellExperiment <- function(
+# nolint end: cyclocomp_linter
+    adata,
+    assays_mapping = NULL,
+    col_data_mapping = NULL,
+    row_data_mapping = NULL,
+    reduction_mapping = NULL,
+    colPairs_mapping = NULL, # nolint
+    rowPairs_mapping = NULL, # nolint
```
rcannood, line 87: why is `col_data_mapping` snakecase but `colPairs_mapping` camelcase?

vignettes/singlecellexperiment.Rmd
rcannood, line 1: Could this vignette be renamed to `detailed_singlecellexperiment` or something like this?

Also, could you add an entry to `_pkgdown.yml`?

rcannood (approved):

rcannood:
Minor comments, otherwise LGTM!

R/SingleCellExperiment.R
LouiseDck: Good point, no reason and we should be consistent.

vignettes/singlecellexperiment.Rmd
LouiseDck: Yep!

rcannood (approved):

PR #216, opened 2025-01-16
fix issue by explicitly adding dimensions to arrays and vectors

rcannood:
Reading an array now explicitly adds the dimensions to the object to allow functions to check whether the length of an array in obsm is equal to the number of observations in the AnnData, and the same for the length of an array in varm and the number of variables respectively.


LouiseDck (approved):
:+1:

PR #228, opened 2025-02-24
Make `HDF5AnnData$finalize()` private

lazappi:
From v2.6.0 **{R6}** warns about public `finalize()` methods. This PR makes the method private and adds a version to the **{R6}** dependency.


rcannood (approved):
I see, makes sense!

PR #230, opened 2025-02-25
Standardise styling

lazappi:
Add some tools to help standardise styling:

- Add a check for code formatting using [`air`](https://www.tidyverse.org/blog/2025/02/air/) to the `lint` GitHub action
- Add an `air` config file
- Add a check for vignette code formatting using **{styler}** to the `lint` GitHub action
- Restyle the package

I have mainly `air` because it is fast and has some nicer (to me) defaults (especially for some function definitions). On the other hand, it is new so the styling might be a bit unstable while it is being developed. It also can't be used for `.Rmd` files which is why I have used **{styler}** to check those. I'm happy to switch to just **{styler}** if we think that is easier.

Side notes:

- Bioconductor styler is 4 space indentation so we might need to change to that at some point
- I read some pretty good arguments for why tab indentation is better than spaces for accessibility which I hadn't considered before. Not saying we should switch now, but just FYI.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -6,6 +6,7 @@
 ^\.metals$
 ^\.vscode$
 ^\.lintr$
+^\.air.toml$
 ^_pkgdown\.yml$
 ^docs$
 ^pkgdown$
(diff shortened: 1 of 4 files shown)


rcannood (approved):
LGTM!

PR #231, opened 2025-02-26
Use  {cli} for alerts

lazappi:
We already depend on **{cli}** so I think it makes sense to use it for errors, warnings etc. It should help give more informative messages and makes some of the code nicer.

I think some of the checks etc. can be refactored so that code isn't repeated but I'll leave that for a separate PR.


R/AbstractAnnData.R
```diff
@@ -317,25 +324,59 @@ AbstractAnnData <- R6::R6Class(
...
-          label,
-          ". Expected: [",
-          paste(shape, collapse = ", "),
-          "], found: [",
-          paste(mat_dims, collapse = ", "),
-          "]."
-        ))
+        cli_abort(
+          c(
+            "Unexpectd shape for {.field {label}}",
```
rcannood, line 329:
```suggestion
            "Unexpected shape for {.field {label}}",

File: R/AbstractAnnData.R [hunk 2]
```diff
@@ -317,25 +324,59 @@ AbstractAnnData <- R6::R6Class(
...
+            "i" = paste0(
+              "Expected [",
+              paste(shape, collapse = ", "),
+              "] got [",
```
rcannood, line 333:
```suggestion
              "], but got [",

rcannood (commented):
Much better!

Some minor comments, still need to review a few files.

rcannood (approved):
Great work. Except for the two minor suggestions, LGTM!

PR #237, opened 2025-03-06
Bugfix: assay_metadata issue when going AD -> Seurat -> AD -> Seurat

LouiseDck:
When looking into #236 I encountered the following bug:

```{r}
> ad <- generate_dataset(n_obs = 10L, n_var = 20L, format = "AnnData")
> ad$var <- ad$var[c("factor", "logical")]

> so1 <- to_Seurat(
    ad,
    graph_mapping = list(
      numeric_matrix = "numeric_matrix",
      integer_matrix = "integer_matrix"
    )
  )

> ad2 <- from_Seurat(
    so1,
    assay_name = "RNA",
    output_class = "InMemoryAnnData",
    x_mapping = "counts"
  )
> so2 <- to_Seurat(ad2)
Error in names(v) <- row.names(value) : 
  'names' attribute [20] must be the same length as the vector [1]
```

This seems(?) to be due to rownames of the `ad$var` slot?

Fixed by not copying the whole slot.

### DIFF
--- a/tests/testthat/test-Seurat.R
+++ b/tests/testthat/test-Seurat.R
@@ -185,7 +185,7 @@ for (obs_key in colnames(obj@meta.data)) {
     expect_true(obs_key %in% colnames(ad$obs))
     expect_equal(
       ad$obs[[obs_key]],
-      obj[[obs_key]],
+      obj@meta.data[[obs_key]],
       info = paste0("obs_key: ", obs_key)
     )
   })
@@ -204,7 +204,7 @@ for (var_key in colnames(active_assay@meta.data)) {
     skip_if(!is.null(msg), message = msg)
 
     expect_true(var_key %in% colnames(ad$var))
-    expect_equal(ad$var[[var_key]], active_assay[[var_key]])
+    expect_equal(ad$var[[var_key]], active_assay@meta.data[[var_key]])
   })
 }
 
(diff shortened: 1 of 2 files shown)


LouiseDck:
I didn't quite like the solution, this is nicer and simpler I think.

I still don't completely understand the issue, it seems that when using `from_Seurat` the internal (?) structure of the dataframes got changed? I'm not well versed enough in R to know if this can even happen and how it does, but this lead to [this line](https://github.com/satijalab/seurat-object/blob/d4d0db84ae8626a4faa42646a018703edf09edad/R/assay5.R#L2892) failing.
The issue was that e.g. `ad$obs[["factor"]]` produces a factor, but after to_Seurat and from_Seurat, `ad$obs[["factor"]]` still produced a dataframe.

It seems fixed now though.

tests/testthat/test-Seurat.R
```diff
@@ -185,7 +185,7 @@ for (obs_key in colnames(obj@meta.data)) {
     expect_true(obs_key %in% colnames(ad$obs))
     expect_equal(
       ad$obs[[obs_key]],
-      obj[[obs_key]],
```
LouiseDck, line 188: I'm very sorry @lazappi but using the accessors don't seem to extract the vector from the dataframe, which causes issues

rcannood (approved):
LGTM

PR #243, opened 2025-03-11
Add helper for styling vectors in messages

lazappi:
Improve the styling of vectors in error messages:

- Add a `style_vec()` function that can be reused (and makes the arguments more obvious)
- Replace previous uses of `cli::cli_vec()` with `style_vec()`


rcannood (approved):
LGTM!

PR #245, opened 2025-03-11
Pass messages instead of error objects to `cli_abort()`

### LINKED ISSUE #242: Error: Could not evaluate cli `{}` expression: `label`
I updated the package today using  `pak::pak("scverse/anndataR", dependencies = TRUE)` and get the following error? It does not seem obvious to me why I get this error as the package was working fine before I updated it.

```
> adata = from_Seurat(seurat_obj = seuratObj)
Error in `fun(..., .envir = .envir)`:
! Could not evaluate cli `{}` expression: `label`.
Caused by error in `eval(expr, envir = envir)`:
! object 'label' not found
```

```
> packageVersion("anndataR")
[1] ‘0.99.0’
```

lazappi:
Fixes a mistake I introduced when updating the errors to use **{cli}** where `cli_abort()` is called on an error object instead of just a message leading to cryptic errors. Also fixes an extra comma which was making things even more confusing.

Fixes #242

### DIFF
--- a/R/AbstractAnnData.R
+++ b/R/AbstractAnnData.R
@@ -339,7 +339,7 @@ AbstractAnnData <- R6::R6Class(
               "], got [",
               paste(mat_dims, collapse = ", "),
               "]"
-            ),
+            )
           ),
           call = rlang::caller_env()
         )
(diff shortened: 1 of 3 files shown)


rcannood (approved):
LGTM!

PR #246, opened 2025-03-12
Restructure conversion functions

lazappi:
A lot of the code in `from_Seurat()`/`from_SingleCellExperiment()` was inside a big `tryCatch()` block. The made it difficult to identify the source of errors. This PR:

- Modifies `from_Seurat()` and `from_SingleCellExperiment()` to collect slots in a list and delay constructing an `AnnData` until the end of the function, reducing the `tryCatch()` to just the final step
- Added context to the errors from `tryCatch()` to help with identifying problems

While I was making this change I found the code difficult to navigate so I also restructured these functions (this should also help with identifying the source of errors):

- Moved code to internal processing functions instead of the main `from_*` function
- Tidied setting of mappings using `%||%`
- Split the `from_*` and `to_*` functions into separate files
- Split the corresponding tests into separate files


lazappi:
I realise the diffs aren't very helpful because of the new files but I haven't add much new code, mostly moved stuff around

rcannood:
This is indeed hard to review, but I fully agree with the proposed changes! :) 

Given that the test pass, do you want me to check the changes in more detail?

lazappi:
Up to you and @LouiseDck. I'm pretty confident the code still works but everyone should be happy with the structure before it's merged. Maybe worth skimming to see if you're ok with how it's organised?

lazappi:
I re-merged the files to fix the conflicts with `main` so you can see the changes now. We can split them again before this is merged/in a separate PR.

LouiseDck:
As I'm on vacation atm, the earliest I would be able to review the code well would be Friday evening or Saturday.

The proposed changes, as you've written them do sound like good and needed.

rcannood (approved):
Nice, looks great!

PR #232, opened 2025-02-27
Fix missing `NA`s in `SelfHits` known issue

lazappi:
Fix the known issue where `NA` values are converted to 0 values in `rowPairs`/`colPairs` in `to_SingleCellExperiment()`.

This happens because **{SingleCellExperiment}** only keeps values `> 0`. See https://github.com/drisso/SingleCellExperiment/issues/80.

PR #248, opened 2025-03-12
Update known `SelfHits` issue

lazappi:
Update the known issue with `SelfHits` to note that it has been addressed in **{SingleCellExperiment}** and will be fixed in the next Bioconductor release.

Replaces #232 which adds the same fix to **{anndataR}**


rcannood (approved):
LGTM! Thanks!

PR #250, opened 2025-03-13
Use PAT in pr-commands workflow

lazappi:
Currently, the PR commands work and update a PR but the checks are not rerun because that is specifically prohibited when using the default `GITHUB_TOKEN`. This is inconvenient because you can't see if styling etc. worked until something happens to force the checks to run. Switching to a PAT should (hopefully) trigger the checks as for a normal commit.

### DIFF
--- a/.github/workflows/pr-commands.yaml
+++ b/.github/workflows/pr-commands.yaml
@@ -14,15 +14,17 @@ jobs:
     name: document
     runs-on: ubuntu-latest
     env:
-      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+      GITHUB_PAT: ${{ secrets.GTHB_PAT }}
     permissions:
       contents: write
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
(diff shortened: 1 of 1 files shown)


lazappi:
At the moment the commits from the command come from whoever ran it. Do we want that or would we prefer they come from a bot user?

rcannood (approved):
LGTM

.github/workflows/pr-commands.yaml
```diff
@@ -14,15 +14,17 @@ jobs:
     name: document
     runs-on: ubuntu-latest
     env:
-      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+      GITHUB_PAT: ${{ secrets.GTHB_PAT }}
```
rcannood, line 17: Do we need to store the GTHB_PAT in the env if we also pass that info separately to the `pr-fetch` and `pr-push` commands?

lazappi: Maybe, I'm not sure? The template workflow had it that way so I left it.

PR #251, opened 2025-03-14
Simplify to_Seurat interface

rcannood:
I haven't implemented this yet but I created a draft PR so I don't forget I still need to do this


rcannood:
Derp, I didn't want to turn this into an actual PR

PR #253, opened 2025-03-14
Update object documentation

lazappi:
Update the object documentation man pages


R/Seurat.R
LouiseDck, line 1: I'm not entirely convinced about moving the documentation for the as_Seurat and as_SingleCellExperiment functions to a separate file.
I can understand a reasoning for it, but I prefer keeping the documentation close to the code.

LouiseDck (approved):

R/HDF5AnnData.R
```diff
@@ -1,8 +1,13 @@
...
 #'
 #' @description
-#' Implementation of an in memory AnnData object.
-#' @noRd
+#' Implementation of an HDF5-backed `AnnData` object.
+#'
+#' See [AnnData-usage] for details on creating and using `AnnData` objects.
+#'
+#' @seealso [AnnData-usage] for details on creating and using `AnnData` objects
+#'
```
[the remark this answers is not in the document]
lazappi: `@seealso` is a separate section at the end of the man page. Possibly it's not needed is this case.

R/Seurat.R
lazappi: Yeah, I thought it was a good idea then after I did it I wasn't sure anymore. I'm happy to move it back but then it would make sense to rename `from_Seurat()` and `from_SingleCellExperiment()`. I still want to split `from`/`to` into separate files once we are done with PRs.

R/AnnData.R
```diff
@@ -76,3 +60,240 @@ AnnData <- function(
...
+#'
+#' sce <- SingleCellExperiment(
+#'   assays = list(counts = matrix(1:5, 5L, 3L)),
+#'   colData = DataFrame(cell = 1:3, row.names = paste0("Cell", 1:3)),
+#'   rowData = DataFrame(gene = 1:5, row.names = paste0("Gene", 1:5))
+#' )
+#'
+#' as_AnnData(sce)
+# nolint start: object_name_linter
+as_AnnData <- function(
```
rcannood, line 215: Maybe move this to `as_AnnData.R`?

rcannood (approved):
The updated documentation looks excellent!

Indeed, I think I'd expect roxygen docs to be next to the function it documents in general.

However, I'd like to get this merged asap, since it's a big improvement over the current docs, and also a big change to the codebase.

What do you all think?

PR #255, opened 2025-03-31
Don't assume X is counts or data

LouiseDck:


rcannood:
Could you merge main into your branch? :bow:

lazappi:
@LouiseDck I think `main` can be merged here now. I looked at doing it but I wasn't sure how to resolve the conflicts and I didn't want to break anything.

R/Seurat.R
```diff
@@ -37,9 +39,8 @@
 #' If `NULL`, the internal function `.to_Seurat_guess_layers` will be used to
 #' guess the layer mapping as follows:
 #'
-#' * All AnnData layers are copied to Seurat layers by name
-#' * If there is layer named "counts", `adata$X` is stored a "data" otherwise
-#'   it is stored as "counts"
+#' * All AnnData layers are copied to Seurat layers by name.
+#' * adata$X will not be copied to Seurat layers, unless an `x_mapping` is provided.
```
lazappi, line 43: Do we think this is the correct default? For other slots we usually try to copy as much stuff as possible so I would probably copy it to a layer named `"X"` by default. Maybe this makes sense though.

R/Seurat.R
```diff
@@ -179,16 +181,33 @@ to_Seurat <- function(
...
-  # check seurat layers
-  if (is.null(names(layers_mapping))) {
-    names(layers_mapping) <- layers_mapping
+  # check seurat layers (which includes the X mapping)
+  if (!is.null(x_mapping)) {
+    if (any(is.na(layers_mapping))) {
+      cli_abort(
+        "{.arg layers_mapping} must not contain any {.val NA} values when {.arg x_mapping} is provided"
+      )
+    }
```
lazappi, line 190: Maybe we don't need to support `NA` in `layers_mapping` at all any more not that there is an `x_mapping` argument? It might be simpler to only have one way of defining how to map `X`.

R/Seurat.R
```diff
@@ -508,25 +532,48 @@ to_Seurat <- function(
...
 
 # nolint start: object_name_linter object_length_linter
-.to_Seurat_guess_layers <- function(adata) {
+.to_Seurat_guess_layers <- function(adata, x_mapping) {
   # nolint end: object_name_linter object_length_linter
   layers <- self_name(adata$layers_keys())
+  if (!is.null(adata$X) && is.null(x_mapping)) {
+    layers[["X"]] <- NA
+  }
+  layers
```
lazappi, line 541: Ah, we do convert `X` by default here. Maybe the docs above just need to be updated then. There probably needs to specify not converting `X`, I guess `x_mapping = c()` would be consistent with other arguments.

R/Seurat.R
```diff
@@ -508,25 +532,48 @@ to_Seurat <- function(
...
-        "counts"
-      } else {
-        "data"
-      }
+# nolint start: object_name_linter object_length_linter
+.to_Seurat_guess_reductions <- function(adata) {
+  # nolint end: object_name_linter object_length_linter
+  if (!inherits(adata, "AbstractAnnData")) {
+    "{.arg adata} must be a {.cls AbstractAnnData} but has class {.cls {class(adata)}}"
+  }
```
lazappi, line 549: I don't think we need this check if it's in the user function and we only use this internally.

R/Seurat.R
```diff
@@ -508,25 +532,48 @@ to_Seurat <- function(
...
+
+  reductions <- list()
 
-    layers[layer_name_for_x] <- NA
+  for (reduction_name in names(adata$obsm)) {
+    if (grepl("^X_", reduction_name)) {
```
lazappi, line 554: Why check for `X_`, shouldn't we convert any reductions regardless of name?

R/Seurat.R
```diff
@@ -508,25 +532,48 @@ to_Seurat <- function(
...
+      out <-
+        if (reduction_name == "X_pca" && "PCs" %in% names(adata$varm)) {
+          list(key = "X_pca", obsm = "X_pca", varm = "PCs")
```
lazappi, line 557: This is probably a reasonable special case as long as it's documented.

R/SingleCellExperiment.R
```diff
@@ -196,7 +218,7 @@ to_SingleCellExperiment <- function(
 }
 
 # nolint start: object_length_linter object_name_linter
-.to_SCE_guess_assays <- function(adata) {
+.to_SCE_guess_assays <- function(adata, x_mapping) {
```
lazappi, line 221: I don't think this is used above any more so maybe it can be removed?

vignettes/seurat_mapping.Rmd
```diff
@@ -123,12 +137,14 @@ seurat_object <- to_Seurat(
...
-### layers_mapping
+### layers_mapping and x_mapping
 
 Seurat allows for multiple layers to be stored in the object. This corresponds with the `layers` slot in the AnnData object.
-If no mapping is provided, we will try to guess which layer the `X` anndata slot corresponds to by checking whether another
-layer called `counts` or `data` exists in the AnnData object.
-The other layers of the AnnData object will be copied by name.
+If no mapping is provided, all layers of the AnnData object will be copied by name. The `X` slot will also be copied by name.
+
+There must be layer called `counts` or `data` in the final Seurat object. If no AnnData layers are named `counts` or `data`, you must
```
lazappi, line 145:
```suggestion
There must be a layer called `counts` and/or `data` in the final Seurat object. If no AnnData layers are named `counts` or `data`, you must
```

lazappi (changes requested):
I made a few comments. Mostly to clarify how things are supposed to work.

R/Seurat.R
LouiseDck: Oh yeah, this whole function is a result of the quite complicated merge with main :thinking: This function is duplicated, and I think this is just an old version of it. The actual good version of the function is below here. Thanks for catching it, this happened with quite a few functions and I thought I got them all.

R/Seurat.R
LouiseDck: See above, this is still the case in the 'good' version.

R/Seurat.R
LouiseDck: idem, see longer explanation in other comment.

R/Seurat.R
LouiseDck: It is simpler, but `x_mapping` is a special-case shortcut in the way it works. I think it makes sense to still allow people to map `X` using `layers_mapping`, as that follows the general idea of the mapping arguments still.

R/Seurat.R
LouiseDck: Good catch, no, it is copied by default to 'X'. I'll change the documentation.

R/Seurat.R
LouiseDck: Hmm, I do think that by default we want it converted the same way as the layers. If the user does not want it converted, they can specify this by not providing `x_mapping` and not including `X` in `layers_mapping`

R/SingleCellExperiment.R
LouiseDck: Ah yeah, my bad.

tests/testthat/test-SingleCellExperiment.R
```diff
@@ -85,6 +85,41 @@ for (layer_key in names(ad$layers)) {
...
+      x_mapping = "counts",
+      assays_mapping = c(counts = "numeric_matrix", integer = "integer_matrix")
+    ),
+    regexp = "duplicate names"
+  )
+})
+
+test_that("to_SCE works when only providing x_mapping", {
+  sce <- ad$as_SingleCellExperiment(x_mapping = "counts")
+  assay_names <- names(assays(sce))
```
lazappi, line 100: There is a function to get the names:

```suggestion
  assay_names <- assayNames(sce)
```

R/Seurat.R
```diff
@@ -171,16 +210,22 @@ as_Seurat <- function(
...
+  )
+
   obj <- SeuratObject::CreateSeuratObject(
+    assay,
     meta.data = object_metadata,
-    assay = assay_name,
-    counts = counts,
-    data = data
+    assay = assay_name
   )
```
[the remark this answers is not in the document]
LouiseDck: IIRC, it threw up extra warnings when trying to create a SeuratObject not containing a `counts` slot, or if the `counts` slot was `NULL`.

PR #260, opened 2025-04-14
Simplify mapping arguments

### LINKED ISSUE #239: `from_Seurat`: Simplify varm_mapping and other functions to length 1

lazappi:
Simplify the mapping arguments in conversion functions to use named vectors instead of nested lists.

Related to (replaces?) #251, fixes #239

- All mapping arguments now expect named vectors instead of lists
- Complex mapping from other slots etc. is now not possible. Users will need to make these changes manually if required.
- The exception is the `reduction_mapping` argument of `from_Seurat()`/`from_SingleCellExperiment()` which allows a more complex list format if users want to specify loadings etc.
- Simple named lists will still work but are not shown in examples
- Unnamed vectors will also work, `c("item1", "item2")` gets converted to `c(item1 = "item1", item2 = "item2")`. This is mostly to avoid simple errors and isn't shown in examples.

This helps standardise things so it will be simpler to replace `from_Seurat()`/`from_SingleCellExperiment()` with `as_AnnData()`. More updates to documentation will come in #253.


R/SingleCellExperiment.R
```diff
@@ -14,48 +14,61 @@
...
-#' metadata names in the created SingleCellExperiment object.
-#' The names of the list should be the names of the `metadata` in the resulting
-#' SingleCellExperiment object. The values of the list should be the names of the
-#' `uns` in `adata`.
+#' @param assays_mapping A named vector mapping `layers` in `adata` to `assay`
+#'   names in the created SingleCellExperiment object. The names of the vector
+#'   should be the names of the `assays` in the resulting SingleCellExperiment
+#'   object, and the values should be the names of the `layers` in `adata`, and
+#'   can include the `X` matrix as well. If `X` is not in the list, it will be
+#'   added as `counts` or `data`.
```
LouiseDck, line 22: I think this should be more specific: will it be added as `counts` or `data`? In what cases as `counts` or as `data`?

LouiseDck (approved):
Looks good to me!

R/SingleCellExperiment.R
lazappi: I was trying not to change this here because I thought you would look at it in #255. But I can do it here if that's easier.

R/Seurat.R
```diff
@@ -918,28 +947,14 @@ from_Seurat <- function(
     return(invisible())
```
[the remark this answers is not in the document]
lazappi: These functions don't return anything, they just modify `adata` in place. A normal `return()` returns `NULL` which might be printed to something, `invisible()` makes sure the result is hidden. Possibly there is a nicer way to do this.

lazappi:
I'm going to merge this because I think it's holding up other things. I have another PR which should sort out any fixes to the documentation.

R/SingleCellExperiment.R
LouiseDck: No, indeed it will be clarified in #255

PR #273, opened 2025-05-26
Move `as_*` documentation to functions

### LINKED ISSUE #269: Move documentation according to comments on PR #253
Move documentation back to `SingleCellExperiment.R` and `Seurat.R`.

lazappi:
Move the documentation for `as_*` functions from `as.R` to the function definitions. The internal `to_*` functions are renamed to `as_*`.

Fixes #269


rcannood (changes requested):
Does it make sense to add a deprecation message for the people already using the to_ functions 

Also, could you add an entry to the changelog?

lazappi:
> Does it make sense to add a deprecation message for the people already using the to_ functions

Possibly. I can go back and check which were exported and add something for those.

> Also, could you add an entry to the changelog?

:+1:

lazappi:
> Does it make sense to add a deprecation message for the people already using the to_ functions

Everything that was exported before #253 is now exported and deprecated. I think that's fine for now but I would remove them as soon as we think it's reasonable.

> Also, could you add an entry to the changelog?

I don't think we have a changelog? There's a `NEWS` but we haven't been keeping that up to date.

PR #279, opened 2025-06-06
Split conversion files into to/from

lazappi:
Split the files for conversion functions into separate to/from files to make them easier to manage


lazappi:
How's this? The renamed `as_Seurat()`/`as_SCE()` and moved `as_AnnData()` to a separate file. The name of the `from_*` helpers is harder but we could also change those if you have a good suggestion.

PR #281, opened 2025-06-16
Prepare v0.1.0 release

lazappi:
Prepare an initial release candidate:

- Update authors
- Set version to `0.1.0`
- Set R dependency to `>= 4.1.0` to address check NOTE
- Update NEWS

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,27 +1,9 @@
-# anndataR 0.99.0
+# anndataR 0.1.0 (inital release candidate)
 
-## New features
+Initial release candidate of **{anndataR}** including:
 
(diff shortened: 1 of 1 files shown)


R/AbstractAnnData.R
```diff
@@ -213,7 +213,7 @@ AbstractAnnData <- R6::R6Class(
     #' @return A `SingleCellExperiment` object
     to_SingleCellExperiment = function(...) {
       lifecycle::deprecate_warn(
-        "0.99.0",
+        "0.1.0",
```
rcannood, line 216: Shouldn't this be 0.2.0 or 0.99 (that is, when will it be deprecated & removed?)

rcannood (changes requested):
Almost lgtm!

Issue #275, opened 2025-06-02
Bioconductor submission pre-review


lazappi:
This is a pre-review of the package to make the submission process smoother.
We should try to address as many of these as possible before submitting.
It might make sense to create sub-issues for some of these that are bigger tasks and/or need discussion.
Try to keep this checklist up to date as we fix things.

# Review

**Key:** :rotating_light: Required :warning: Recommended :green_circle: Optional :question: Question

_My comments/suggestions/explanations are in italics_

## General package development

- [ ] :rotating_light: Fix the `NOTE`s from `R CMD build`

    ```
    WARNING: Added dependency on R >= 4.1.0 because package code uses the
     pipe |> or function shorthand \(...) syntax added in R 4.1.0.
     File(s) using such syntax:
       ‘Seurat.R’ ‘SingleCellExperiment.R’
    checking for unstated dependencies in vignettes ... NOTE
      'library' or 'require' calls not declared from:
        ‘dplyr’ ‘rprojroot’ ‘stringr’ ‘tibble’ ‘tidyr’
    ```
- [ ] :warning: Fix `NOTE`s and `WARNING`s from `BiocCheck::BiocCheck(`new-package` = TRUE)`

  ```
  ℹ NOTE: Update R version dependency from 4.0.0 to 4.5.0
  ! WARNING: No Bioconductor dependencies detected. Note that some infrastructure packages may not have Bioconductor dependencies. For more information, reach out to the Bioconductor community and/or consider a CRAN submission.
  ℹ NOTE: 'sessionInfo' not found in vignette(s)
  ℹ NOTE: Avoid sapply(); use vapply()
  ℹ NOTE: Avoid 'cat' and 'print' outside of 'show' methods
  ! WARNING: .Deprecated / .Defunct usage (found 5 times)
  ℹ NOTE: Use accessors; don't access S4 class slots via '@' in examples/vignettes.
  ℹ NOTE: Avoid 'suppressWarnings'/'*Messages' if possible (found 3 times)
  ℹ NOTE: The recommended function length is 50 lines or less. There are 14 functions greater than 50 lines.
  ℹ NOTE: Usage of dontrun / donttest tags found in man page examples. 7% of man pages use at least one of these tags.
  ℹ NOTE: Use donttest instead of dontrun.
  ℹ NOTE: Consider shorter lines; 273 lines (4%) are > 80 characters long.
  ℹ NOTE: Consider multiples of 4 spaces for line indents; 2342 lines (31%) are not.
[...]

PR #291, opened 2025-07-06
Use accessors

LouiseDck:
**Related to:** #275 

- [x] :rotating_light: Use accessors instead of `@` or `slot()` (where possible)

  _There was a note about this so we should check, probably in the `Seurat` stuff_

## Description

Removes direct slot access, where possible.

It was not directly possible in 2 instances:

1. `write_h5ad_helpers.R`, line 262-263

```
  write_h5ad_dense_array(value@p, file, paste0(name, "/indptr"), compression)
  write_h5ad_dense_array(value@x, file, paste0(name, "/data"), compression)
```
In a previous line, the code `attr(value, "i")` is used which we could do here, but I don't think that is meaningfully different.

3. `test-as_Seurat.R`, line 42
We use it to compare cell-level metadata.
```
seu@meta.data[[obs_key]],
``` 
Changing this to `seu[[obs_key]]` does not work as it will find the dimensionality reduction with the same name, and not the cell-level metadata.
Small example to show this:
```
data <- generate_dataset(
  format = "AnnData",
  n_obs = 10,
  n_var = 20
)

seu <- ad$as_Seurat(x_mapping = "counts")

seu[["integer"]]
seu@meta.data[["integer"]]
Embeddings(seu, "integer")
```

## Checklist

**Before review**

- [x] Add/update tests
- [x] Pass CI checks

**Before merge**

- [x] Update `NEWS`
- [x] Bump devel version


R/write_h5ad_helpers.R
```diff
@@ -259,8 +259,8 @@ write_h5ad_sparse_array <- function(
     paste0(name, "/indices"),
     compression
   )
-  write_h5ad_dense_array(value@p, file, paste0(name, "/indptr"), compression)
-  write_h5ad_dense_array(value@x, file, paste0(name, "/data"), compression)
+  write_h5ad_dense_array(value@p, file, paste0(name, "/indptr"), compression) # accessor here
+  write_h5ad_dense_array(value@x, file, paste0(name, "/data"), compression) # accessor here
```
LouiseDck, line 263: On these lines we access the slots directly.

tests/testthat/test-as_Seurat.R
```diff
@@ -38,11 +38,12 @@ for (obs_key in colnames(ad$obs)) {
     )
     skip_if(!is.null(msg), message = msg)
 
-    expect_true(obs_key %in% colnames(seu@meta.data))
+    expect_true(obs_key %in% colnames(seu[[]]))
     expect_equal(
-      seu@meta.data[[obs_key]],
+      seu@meta.data[[obs_key]], # accessor here
```
LouiseDck, line 43: Here we still access Seurat internals directly.

lazappi:
```suggestion
      seu[[]][[obs_key]],
```

R/write_h5ad_helpers.R
lazappi: I don't think there is a way around these, so let's just leave them

lazappi, line 263:
```suggestion
  write_h5ad_dense_array(value@p, file, paste0(name, "/indptr"), compression)
  write_h5ad_dense_array(value@x, file, paste0(name, "/data"), compression)
```

lazappi (approved):

PR #294, opened 2025-07-07
Change sapply to vapply

LouiseDck:
**Related to:** #275 and #293 

## Description

Most uses of sapply happened in the roundtrip tests. When running them, they failed due to invalidated file ids. Some gc() fixed this in #293, but I presume they're not being ran in the github actions.

## Checklist

**Before review**

- [x] Add/update tests
- [x] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: anndataR
 Title: AnnData interoperability in R
-Version: 0.1.0.9000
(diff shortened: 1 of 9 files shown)


lazappi (approved):
I'm not sure if you are done with this but looks good to me.

PR #322, opened 2025-08-23
Add reticulate class

rcannood:
**Related to:** 

## Description

I implemented a ReticulateAnnData class so you don't need to write to disk to convert an AnnData object from an h5mu file to SCE/Seurat. See vignette for more an example.

## changelog

* Remove CRAN anndata as a dependency (not sure why this was still the case)
* Add ReticulateAnnData class
* Add tests
* Add vignette

* We could add a `as_AnnData.anndata._core.anndata.AnnData` as well, but I feel like it would be a bit redundant with `reticulate::py_to_r` and it would have a weird function signature (x, x_mapping, layers_mapping, obs_mapping, ...).

## Checklist

**Before review**

- [x] Add/update tests
- [x] Pass CI checks

**Before merge**

- [x] Update `NEWS`
- [ ] Bump devel version


rcannood:
Based on our meeting:

* @lazappi isn't super stoked about adding this backend conceptually because we're trying to promote using our native reader.
* That makes sense, but there are some use cases we can't otherwise cover (e.g. converting mudata to SCE)
* We need to add some info to the docs to explain when to use which backend and why
* This would allow us to deprecate dynverse/anndata on CRAN

.github/workflows/pkgdown.yaml
```diff
@@ -40,6 +40,12 @@ jobs:
           extra-packages: local::.
           needs: website
 
+      - name: Install Python dependencies for tests and vignettes
+        run: |
+          library(reticulate)
+          py_install(c("scanpy", "mudata", "dummy_anndata"))
+        shell: Rscript {0}
```
lazappi: I don't think we need this but we might need to add a `py_require("mudata")` somewhere

R/AnnData-usage.R
```diff
@@ -10,7 +10,9 @@
 #'
 #' There are different back ends for `AnnData` objects that inherit from the
 #' abstract [AbstractAnnData] class. For example, the [InMemoryAnnData] stores
-#' data in memory or the [HDF5AnnData] is backed by a H5AD file.
+#' data in memory, the [HDF5AnnData] is backed by a H5AD file, and the
+#' [ReticulateAnnData] provides direct integration with Python AnnData objects
```
lazappi:
```suggestion
#' [ReticulateAnnData] provides direct integration with Python `AnnData` objects
```

R/check_requires.R
```diff
@@ -6,44 +6,137 @@
...
+  # First check if reticulate is available
+  if (!requireNamespace("reticulate", quietly = TRUE)) {
+    cli_abort(
+      c(
+        "{what} requires the {.pkg reticulate} package",
+        "i" = "Install it using {.code install.packages(\"reticulate\")}"
+      ),
+      call = caller_env()
+    )
+  }
```
lazappi: This can probably use `check_requires()`

R/check_requires.R
```diff
@@ -6,44 +6,137 @@
...
-      Bioc = "install.packages(\"BiocManager\"); BiocManager::install"
+    missing_str <- format_python_package_list(missing)
+    # nolint end object_usage_linter
+
+    cli_abort(
+      c(
+        "{what} requires the Python {.pkg {missing}} package{?s}",
+        "i" = paste(
+          "To continue, install {cli::qty(missing)}{?it/them} using",
+          "{.code reticulate::py_install({missing_str})}"
```
lazappi:
```suggestion
          "To continue, require {cli::qty(missing)}{?it/them} using",
          "{.code reticulate::py_require({missing_str})} or install",
          "{cli::qty(missing)}{?it/them} them in your Python environment",
          "using another method" 
```

R/check_requires.R
```diff
@@ -6,44 +6,137 @@
...
+#' @param packages Character vector of R package names
+#' @return Formatted string for use in install.packages()
+#' @noRd
+format_r_package_list <- function(packages) {
+  missing_str <- paste0("\"", paste(packages, collapse = "\", \""), "\"")
+  if (length(packages) > 1) {
+    missing_str <- paste0("c(", missing_str, ")")
+  }
+  missing_str
+}
```
lazappi: Do these need to be different?

R/check_requires.R
```diff
@@ -6,44 +6,137 @@
...
+    install_fun <- get_r_install_function(where)
     # nolint end object_usage_linter
 
     cli_abort(
       c(
         "{what} requires the {.pkg {missing}} package{?s}",
         "i" = paste(
           "To continue, install {cli::qty(missing)}{?it/them} using",
-          "{.code {fun}({missing_str})}"
+          "{.code {install_fun}({missing_str})}"
```
lazappi: This could maybe be runnable?

```suggestion
          "{.run {install_fun}({missing_str})}"
```

R/reticulate_helpers.R
```diff
@@ -0,0 +1,186 @@
...
+#' py_mat <- try_r_to_py(mat)
+#'
+#' # For AnnData objects, prefer the automatic method:
+#' adata <- AnnData(X = mat)
+#' py_adata <- r_to_py(adata)  # Automatically converts
+#' }
+try_r_to_py <- function(x, ...) {
+  if (!requireNamespace("reticulate", quietly = TRUE)) {
+    cli_abort("The {.pkg reticulate} package is required")
+  }
```
lazappi:
```suggestion
  check_requires("Converting R to Python", "reticulate")
```

R/reticulate_helpers.R
```diff
@@ -0,0 +1,186 @@
...
+    reticulate::r_to_py(x)
+  } else if (is.matrix(x) || inherits(x, "Matrix")) {
+    # Handle matrices - may need special treatment for sparse matrices
+    if (inherits(x, "dgCMatrix") || inherits(x, "dgRMatrix")) {
+      # For sparse matrices, convert directly using reticulate
+      reticulate::r_to_py(x)
+    } else {
+      # Regular dense matrix
+      reticulate::r_to_py(x)
+    }
```
lazappi: Do we need these cases if they are all the same?

R/reticulate_helpers.R
```diff
@@ -0,0 +1,186 @@
...
+    for (name in names(x)) {
+      py_dict[[name]] <- try_r_to_py(x[[name]], ...)
+    }
+    py_dict
+  } else if (is.list(x)) {
+    # Convert unnamed list to Python dict if empty, otherwise to Python list
+    if (length(x) == 0) {
+      # Empty list should be a dict for AnnData compatibility
+      reticulate::dict()
+    } else {
```
lazappi: I'm not sure about this. I would just do named lists to `dict` and unnamed lists to `list`. You can make an empty named list.

R/reticulate_helpers.R
```diff
@@ -0,0 +1,186 @@
...
+      # Non-empty unnamed list becomes Python list
+      reticulate::r_to_py(x, ...)
```
lazappi: Do the items here need to be converted individually?

R/reticulate_helpers.R
```diff
@@ -0,0 +1,186 @@
...
+#' # For AnnData objects, prefer the automatic method:
+#' library(reticulate)
+#' ad <- import("anndata", convert = FALSE)
+#' py_adata <- ad$AnnData(X = py_mat)
+#' r_adata <- py_to_r(py_adata)  # Automatically converts to ReticulateAnnData
+#' }
+try_py_to_r <- function(x, ...) {
+  if (!requireNamespace("reticulate", quietly = TRUE)) {
+    cli_abort("The {.pkg reticulate} package is required")
+  }
```
lazappi:
```suggestion
  check_requires("Converting Python to R", "reticulate")
```

R/reticulate_helpers.R
```diff
@@ -0,0 +1,186 @@
...
+  # Handle different Python object types appropriately for AnnData
+  if (inherits(x, "pandas.core.frame.DataFrame")) {
+    # Convert pandas DataFrame to R data.frame
+    reticulate::py_to_r(x)
+  } else if (inherits(x, "numpy.ndarray")) {
+    # Convert numpy array to R matrix/array
+    reticulate::py_to_r(x)
+  } else if (inherits(x, "scipy.sparse.base.spmatrix")) {
+    # Convert scipy sparse matrix to R sparse matrix
+    reticulate::py_to_r(x)
```
lazappi: Do we need these cases if they are all the default?

R/reticulate_helpers.R
```diff
@@ -0,0 +1,186 @@
...
+  } else {
+    # Handle AnnData special mapping objects (layers, obsm, varm, etc.)
+    if (grepl("anndata.*mapping", paste(class(x), collapse = " "))) {
+      tryCatch(
+        {
+          # Convert to regular dict first, then to R list
+          py_dict <- reticulate::py_eval("dict")(x)
```
lazappi: I usually do:

```r
py_builtins <- reticulate::import_builtins()
py_builtins$dict(x)
```

But this also works

R/reticulate_helpers.R
```diff
@@ -0,0 +1,186 @@
...
+          return(reticulate::py_to_r(py_dict))
+        },
+        error = function(e) {
+          # Fallback: create empty list
+          list()
```
lazappi: I'm not sure about this, or at least there should be a warning

R/reticulate_helpers.R
```diff
@@ -0,0 +1,186 @@
...
+    # Convert scipy sparse matrix to R sparse matrix
+    reticulate::py_to_r(x)
+  } else {
+    # Handle AnnData special mapping objects (layers, obsm, varm, etc.)
+    if (grepl("anndata.*mapping", paste(class(x), collapse = " "))) {
+      tryCatch(
+        {
+          # Convert to regular dict first, then to R list
+          py_dict <- reticulate::py_eval("dict")(x)
+          return(reticulate::py_to_r(py_dict))
```
lazappi: Should individual items be converted here?

R/reticulate_helpers.R
```diff
@@ -0,0 +1,186 @@
...
+    if (inherits(x, "python.builtin.dict")) {
+      # Convert Python dict to R named list using bi$list() for keys
+      bi <- reticulate::import_builtins()
+      dict_keys <- reticulate::py_to_r(bi$list(x$keys()))
+      result <- list()
+      for (key in dict_keys) {
+        result[[key]] <- try_py_to_r(x[[key]], ...)
+      }
+      result
+    } else {
```
lazappi: There are a few converters I needed to add to **{zellkonverter}** https://github.com/theislab/zellkonverter/blob/devel/R/reticulate.R. Maybe worth checking if we need any of these here.

tests/testthat/helper-skip_if_no_anndata_py.R
lazappi, line 1: General question, will this affect the existing round trip tests? I guess they still pass so probably not.

lazappi (changes requested):

tests/testthat/helper-skip_if_no_anndata_py.R
rcannood: anndata wasn't really doing anything anymore at this stage, so no :)

R/check_requires.R
rcannood: No, my bad!

R/reticulate_helpers.R
rcannood: I got rid of the try_py_to_r and try_r_to_py. For some things this used to be necessary but half of the cases here didn't do anything special >_>

R/reticulate_helpers.R
rcannood: See above

R/reticulate_helpers.R
rcannood: See above

R/reticulate_helpers.R
rcannood: See above

R/reticulate_helpers.R
rcannood: :+1:

R/reticulate_helpers.R
rcannood: See above

R/reticulate_helpers.R
rcannood: See above :sweat_smile:

R/reticulate_helpers.R
rcannood: See above :sweat_smile:

.github/workflows/pkgdown.yaml
rcannood: Fixed, thanks!

rcannood:
Thanks for the thorough review! I went through your comments :)

R/ReticulateAnnData.R
```diff
@@ -0,0 +1,446 @@
...
+    var_names = function(value) {
+      private$.check_py_object_valid()
+
+      if (missing(value)) {
+        # trackstatus: class=ReticulateAnnData, feature=get_var_names, status=done
+        bi <- reticulate::import_builtins()
+        reticulate::py_to_r(bi$list(private$.py_anndata$var_names))
+      } else {
+        # trackstatus: class=ReticulateAnnData, feature=set_var_names, status=done
+        private$.py_anndata$var_names <- reticulate::r_to_py(value)
```
lazappi, line 132: Only a couple of these have `reticulate::`. I guess it's imported somewhere?

R/ReticulateAnnData.R
```diff
@@ -0,0 +1,446 @@
...
+          expected_rownames = rownames(self),
+          expected_colnames = colnames(self)
+        )
+        private$.py_anndata$layers <- r_to_py(value)
+        self
+      }
+    },
+
+    #' @field obs See [AnnData-usage]
+    obs = function(value) {
```
lazappi, line 78: I guess we are relying on **{reticulate}** to handle it if someone tries to set part of a slot (like `adata$obs$col <- xx`)? Or does that not work?

Actually, I'm not sure what happens with `HDF5AnnData` either...

vignettes/usage_python.Rmd
lazappi, line 1: There are some formatting things I would adjust here but I'm going to review the vignettes anyway so I'll do it then (assuming this is merged already).

NEWS.md
```diff
@@ -4,6 +4,7 @@
 
 * Implemented an `AnnDataView` class, which provides a lazy view of an `AnnData` object without copying data (PR #1096)
 * Implemented S3 methods for `AbstractAnnData` objects: `dim`, `nrow`, `ncol`, `dimnames`, `rownames`, `colnames`, and `[` (PR #1096)
+* Add ReticulateAnnData class for seamless Python integration via reticulate (PR #322)
```
lazappi:
```suggestion
* Add `ReticulateAnnData` class for seamless Python integration via **{reticulate}** (PR #322)
```

lazappi (approved):
I think this is looking ok. I'm still hoping it doesn't get used much but there are probably some specific use cases where it's nice to have.

R/ReticulateAnnData.R
rcannood: Indeed -- we need to import it to get the `py_to_r.anndata._core.AnnData` (or sth like this) to work properly.

R/ReticulateAnnData.R
rcannood: When we call `adata$obs$col <- "foobar"`, R will:

1. fetch the obs from adata with reticulate
2. set obs$col <- "foobar" in R
3. set adata$obs <- new_obs with reticulate

It's my understanding that when calling something like `a$b$c$d <- a$b$c$d + 10`, this will essentially run something like this:

```r
new_d <- a$b$c$d + 10
new_c <- a$b$c
new_c$d <- new_d
new_b <- a$b
new_b$c <- new_c
a$b <- new_b
```

Reprex:

```r
# Define a simple class with custom $ and $<- methods
my_obj <- function(data = list(), name = "obj") {
  structure(data, class = "my_obj", name = name)
}

# S3 method for $ (getter)
`$.my_obj` <- function(x, name) {
  obj_name <- attr(x, "name") %||% "obj"
  cat("GET: ", obj_name, "$", name, "\n", sep = "")
  NextMethod()
}

# S3 method for $<- (setter) 
`$<-.my_obj` <- function(x, name, value) {
  obj_name <- attr(x, "name") %||% "obj"
  value_desc <- if (inherits(value, "my_obj")) {
    paste0("my_obj(", attr(value, "name") %||% "unnamed", ")")
  } else {
    as.character(value)
  }
  cat("SET: ", obj_name, "$", name, " <- ", value_desc, "\n", sep = "")
  result <- NextMethod()
  attr(result, "name") <- attr(x, "name")  # preserve name
  result
}

# Create deeply nested structure: a$b$c$d
a <- my_obj(list(
  b = my_obj(list(
    c = my_obj(list(
      d = 5
    ), name = "c")
  ), name = "b")
), name = "a")


a$b$c$d <- a$b$c$d + 10
#> GET: a$b
#> GET: b$c
#> GET: c$d
#> GET: a$b
#> GET: b$c
#> SET: c$d <- 15
#> SET: b$c <- my_obj(c)
#> SET: a$b <- my_obj(b)
```

PR #323, opened 2025-08-23
install dummy_anndata in CI

rcannood:
**Related to:** 

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -77,6 +77,12 @@ jobs:
           fi
         shell: bash
 
+      - name: Install Python dependencies for tests
+        run: |
+          library(reticulate)
+          py_install("dummy_anndata")
+        shell: Rscript {0}
+
       - uses: r-lib/actions/check-r-package@v2
         with:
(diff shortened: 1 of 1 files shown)


lazappi:
The helper should already do this using `py_require()` https://github.com/scverse/anndataR/blob/devel/tests/testthat/helper-skip_if_no_dummy_anndata.R. Actually, I'm fairly sure the environment you added it to won't be used by the tests.

rcannood:
I see!

Locally the roundtrip tests were getting skipped because dummy anndata wasn't installed. But maybe I was doing something wrong. Let me verify what happened from my side.

lazappi:
I checked they are running on GHA. Possibly, locally you are finding a different environment. I usually have to do `Sys.setenv(RETICULATE_USE_MANAGED_VENV="yes")` or **{reticulate}** tries to use something else.

rcannood:
Alright, PEBKAC it is :relaxed:

PR #324, opened 2025-08-23
Implement AbstractAnnData S3 methods and lazy subsetting

rcannood:
**Related to:** #189

## Description

* Implemented AnnDataView class, which provides a lazy view of an AnnData object without copying data (PR #1096).
* Implement S3 methods for AbstractAnnData objects: `dim`, `nrow`, `ncol`, `dimnames`, `rownames`, `colnames`, and `[` (PR #1096).

## Checklist

**Before review**

- [x] Update and regenerate man pages
- [x] Add/update tests
- [x] Add/update examples in vignettes
- [x] Pass CI checks

**Before merge**

- [x] Update `NEWS`
- [ ] Bump devel version

Issue #271, opened 2025-05-20
anndataR ignores outcomes of SCT


LouiseDck:
After running SCT on a Seurat dataset you end up with 2 issues:
- scale.data has less features and can't easily be transferred
- there are these extra "SCTModel"s added to the Seurat object, which don't seem to fit cleanly in any of the anndata slots.

Do we want to do anything with this? Do we want to provide a vignette on how to deal with this?

reprex:
``` r
library(Seurat)
#> Loading required package: SeuratObject
#> Loading required package: sp
#> 
#> Attaching package: 'SeuratObject'
#> The following objects are masked from 'package:base':
#> 
#>     intersect, t
library(anndataR)
library(sctransform)
library(SeuratData)
#> ── Installed datasets ──────────────────────────────── SeuratData v0.2.2.9002 ──
#> ✔ pbmc3k 3.1.4
#> ────────────────────────────────────── Key ─────────────────────────────────────
#> ✔ Dataset loaded successfully
#> ❯ Dataset built with a newer version of Seurat than installed
#> ❓ Unknown version of Seurat installed

InstallData("pbmc3k")
#> Warning: The following packages are already installed and will not be
#> reinstalled: pbmc3k
pbmc <- UpdateSeuratObject(pbmc3k)
#> Validating object structure
#> Updating object slots
#> Ensuring keys are in the proper structure
#> Warning: Assay RNA changing from Assay to Assay
#> Ensuring keys are in the proper structure
#> Ensuring feature names don't have underscores or pipes
#> Updating slots in RNA
#> Validating object structure for Assay 'RNA'
#> Object representation is consistent with the most current Seurat version

# store mitochondrial percentage in object meta data
pbmc <- PercentageFeatureSet(pbmc, pattern = "^MT-", col.name = "percent.mt")

# run sctransform
pbmc <- SCTransform(pbmc, vars.to.regress = "percent.mt", verbose = FALSE)
#> Warning: The `slot` argument of `GetAssayData()` is deprecated as of SeuratObject 5.0.0.
#> ℹ Please use the `layer` argument instead.
#> ℹ The deprecated feature was likely used in the Seurat package.
#>   Please report the issue at <https://github.com/satijalab/seurat/issues>.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
#> `vst.flavor` is set to 'v2' but could not find glmGamPoi installed.
[...]

PR #328, opened 2025-08-28
Refactor obs/var_names handling

rcannood:
**Related to:** 

## Description

#324 is starting to accumulate a lot of different, somewhat related changes. I'm going to split it up a bit to try to make the review process a bit easier.

### Minor changes

- Refactor setter methods in `HDF5AnnData` and `InMemoryAnnData` to use pipe 
  operators for cleaner code
- Add explanatory comments for matrix generation alignment with dummy-anndata 

### Bug fixes

- Add compression parameter to additional write operations in `HDF5AnnData` 
  for consistency
- Directly use `obs_names` and `var_names` properties instead of corresponding
  indirect S3 methods `rownames` and `colnames`
- Fix error message variable name in `.validate_aligned_array()` method 
  (expected_colnames → expected_rownames)

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [x] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/R/InMemoryAnnData.R
+++ b/R/InMemoryAnnData.R
@@ -39,6 +39,8 @@ InMemoryAnnData <- R6::R6Class(
     .layers = NULL,
     .obs = NULL,
     .var = NULL,
+    .obs_names = NULL,
+    .var_names = NULL,
     .obsm = NULL,
     .varm = NULL,
     .obsp = NULL,
@@ -50,15 +52,16 @@ InMemoryAnnData <- R6::R6Class(
     X = function(value) {
       if (missing(value)) {
         # trackstatus: class=InMemoryAnnData, feature=get_X, status=done
-        private$.X
+        private$.X |>
+          private$.add_matrix_dimnames("X")
       } else {
         # trackstatus: class=InMemoryAnnData, feature=set_X, status=done
         private$.X <- private$.validate_aligned_array(
           value,
           "X",
           shape = c(self$n_obs(), self$n_vars()),
-          expected_rownames = self$obs_names,
-          expected_colnames = self$var_names
+          expected_rownames = private$.obs_names,
+          expected_colnames = private$.var_names
         )
         self
       }
@@ -67,15 +70,16 @@ InMemoryAnnData <- R6::R6Class(
     layers = function(value) {
       if (missing(value)) {
         # trackstatus: class=InMemoryAnnData, feature=get_layers, status=done
-        private$.layers
+        private$.layers |>
+          private$.add_mapping_dimnames("layers")
       } else {
         # trackstatus: class=InMemoryAnnData, feature=set_layers, status=done
         private$.layers <- private$.validate_aligned_mapping(
           value,
           "layers",
           c(self$n_obs(), self$n_vars()),
-          expected_rownames = self$obs_names,
-          expected_colnames = self$var_names
+          expected_rownames = private$.obs_names,
+          expected_colnames = private$.var_names
         )
         self
       }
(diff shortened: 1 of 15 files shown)


rcannood:
Checks are failing, let me review

rcannood:
The issue is that we're trying to store the loadings into varm, but the Seurat loadings are subset to the HVG only.

Hence we could fill the matrix with 0s such that non-HVG genes have value 0 and the matrix ordered according to the var_names.

@lazappi @LouiseDck wdyt?

lazappi:
This probably makes sense for the loadings but I would still want a warning to tell people it's happening. There is a similar problem with layers which I think is more complicated in #271.

What changed to make the tests start failing?

rcannood:
A warning would make sense!

The tests started failing because the expected rownames and colnames were being filled with colnames/rownames(self), which evaluates to NULL because these s3 methods don't exist yet.

lazappi:
Hmmmm... I'm still not sure why it didn't fail before though. I'm probably just missing something 😸

rcannood:
The `private$.validate_aligned_array(...)` function throws an error if the  `expected_row/colnames` are not NULL and don't match the observed row/colnames. But we were passing `row/colnames(adata)`, but since `dimnames.AbstractAnnData()` are not implemented, calling `row/colnames(adata)` always just returns `NULL` and hence no check takes place :sweat_smile:

tests/testthat/test-from_Seurat.R
```diff
@@ -191,46 +208,58 @@ test_that("as_AnnData (Seurat) works with v3 Assays", {
...
-  expect_warning(as_AnnData(obj, layers_mapping = NULL), "argument is empty")
-  expect_warning(as_AnnData(obj, layers_mapping = c()), "argument is empty")
+  expect_warning(
+    as_AnnData(obj, varm_mapping = NULL),
+    regexp = "argument is empty"
+  )
+  expect_warning(
+    as_AnnData(obj, varm_mapping = c()),
+    regexp = "argument is empty"
+  )
```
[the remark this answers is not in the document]
rcannood: false positive -- but ideally i'd add a comment to the test to explain why the varm_mapping is there.

R/InMemoryAnnData.R
```diff
@@ -106,36 +122,39 @@ InMemoryAnnData <- R6::R6Class(
...
+        private$.obsm |>
+          private$.add_mapping_dimnames("obsm")
       } else {
         # trackstatus: class=InMemoryAnnData, feature=set_obsm, status=done
         private$.obsm <- private$.validate_aligned_mapping(
           value,
           "obsm",
           c(self$n_obs()),
-          expected_rownames = rownames(self)
+          expected_rownames = private$.obs_names,
```
lazappi: Is it better to use `self$obs_names` here?

tests/testthat/test-from_Seurat.R
```diff
@@ -117,17 +118,30 @@ test_that("as_AnnData (Seurat) retains pca", {
   )
   skip_if(!is.null(msg), message = msg)
 
-  # trackstatus: class=Seurat, feature=test_set_obsm, status=wip
+  # trackstatus: class=Seurat, feature=test_set_obsm, status=done
```
lazappi: I think I left this as `wip` because only the PCA is checked. But happy to change it.

R/InMemoryAnnData.R
rcannood: True -- fixed!

R/from_Seurat.R
```diff
@@ -297,7 +297,23 @@ from_Seurat <- function(
...
     }
-    SeuratObject::Loadings(seurat_obj, .reduction)
+    mat <- SeuratObject::Loadings(seurat_obj, .reduction)
+
+    # NOTE: loadings only contains a subset of the rows that should be in the varm.
+    # hence, expand it to all of the varnames in the adata
+    if (!identical(rownames(mat), adata$var_names)) {
+      cli_warn(c(
+        "Row names of a varm matrix coming from Seurat do not match the var names",
+        "Hence we need to expand the matrix to include all var names."
```
lazappi:
```suggestion
        "Row names of {.code Loadings(seurat_obj, {.val {.reduction})} do not match the expected var names",
        "!" = "The matrix will be expanded to include all var names."
```

lazappi (approved):
Couple of minor things but looks pretty good!

tests/testthat/test-from_Seurat.R
rcannood:
```suggestion
  # trackstatus: class=Seurat, feature=test_set_obsm, status=wip

PR #354, opened 2025-09-23
Clarify `generate_dataset()` usage

### LINKED ISSUE #343: Clarify usage of `generate_dataset()`
Please clarify the purpose of `generate_dataset()`. If it is only for internal testing or demonstrations, it should either not be exported or be explicitly documented as such. See `?scuttle::mockSCE` for a precedent of some simple tweaks that might make its purpose clearer.

lazappi:
**Related to:** Fixes #343 

## Description

Clarify usage of `generate_dataset()`:

- Expand `generate_dataset()` documentation
- Move types to exported and documented list objects
- Add checks for valid type arguments
- Generalise layers created when `format = "Seurat"`

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [x] Update `NEWS`
- [ ] Bump devel version


rcannood:
Hmm, while the new implementation is much cleaner, it's a pity we now have even more mock-related pages in the package :p 

Maybe it wouldn't hurt also adding the word "mock" somewhere at the top of the file. I know we already mention 'generator' but maybe 'mock' might trigger something in people where "generator" and "synthetic" doesn't trigger anything.

Other than that, the updated documentation and code structure is much better!

Wdyt?

lazappi:
I combined the man pages and added "mock" in a couple of places. We could also make `generator_types` a function that returns a list with a `example = TRUE` argument if you think that would be nicer. Or happy to make other changes.

rcannood (approved):
> We could also make generator_types a function that returns a list with a example = TRUE argument if you think that would be nicer.

Sounds like a good idea, but the current implementation is LGTM!

lazappi:
I made it a function, I think that's a bit nicer implementation

rcannood (approved):
LGTM! :relaxed:

PR #369, opened 2025-10-14
Improve handling in conversion from `Seurat`

### LINKED ISSUE #368: Converting `Seurat` fails when graphs don't have a `DefaultAssay`
I encountered a Seurat object with the following content:

    An object of class Seurat 
    73202 features across 342178 samples within 2 assays 
    Active assay: RNA (36601 features, 2000 variable features)
     4 layers present: counts.10x_3p_v3.1, counts.10x_5p_v2, data.10x_3p_v3.1, data.10x_5p_v2
     1 other assay present: sketch
     5 dimensional reductions calculated: pca, harmony, umap, harmony.full, umap.full

---

Calling `as_AnnData()` fails:

```r
> adata <- as_AnnData(data, assay_name = "RNA")
Error in if (SeuratObject::DefaultAssay(graph) != assay_name) { : 
  argument is of length zero
```

On this line: https://github.com/scverse/anndataR/blob/1d9145e009dd2ee46323b7f8b0c670a2bb4da107/R/from_Seurat.R#L464

I can't share the data file, but I notice it has two graphs. I can read the graph just fine, but the assay.used seems to be empty for some reason, thus causing the exception above.

```r
> SeuratObject::Graphs(data)
[1] "sketch_nn"  "sketch_snn"
> gr <- data[["sketch_nn"]]
> gr
A Graph object containing 55331 cells
> gr@assay.used
character(0)
> SeuratObject::DefaultAssay(gr)
NULL
```

I suppose that in this case the graph could be skipped, but it would be nice if as_AnnData didn't throw an error but a warning instead

---

In this case, I'm able to circumvent the issue by turning off the `obsp_mapping` for now until this issue is fixed.

```r
adata <- as_AnnData(data, assay_name = "RNA", obsp_mapping = FALSE)
```

lazappi:
**Related to:** Fixes #368 

## Description

- Handle conversion of graphs that don't have an associated assay
- Add a util functions to check dimensions of objects during conversion
- Check dimensions in `from_Seurat()` and skip if they don't match the `AnnData`
- Adjust documentation of default values in conversion man pages

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/R/as_AnnData.R
+++ b/R/as_AnnData.R
@@ -6,39 +6,31 @@
 #'
 #' @param x The object to convert
 #' @param x_mapping A string specifying the data to map to the `X` slot. If
-#'   `NULL`, no data will be copied to the `X` slot.
+#'   `NULL`, no data will be copied to the `X` slot. See below for details.
 #' @param layers_mapping A named character vector where the names are keys of
 #'   `layers` in the new `AnnData` object and values are the names of items in
-#'   the corresponding slot of `x`. See below for default if `NULL` depending on
-#'   the class of `x`.
+#'   the corresponding slot of `x`. See below for details.
 #' @param obs_mapping A named character vector where the names are names of
 #'   `obs` columns in the new `AnnData` object and values are the names of
-#'   columns in the corresponding slot of `x`. See below for default if `NULL`
-#'   depending on the class of `x`.
+#'   columns in the corresponding slot of `x`. See below for details.
 #' @param var_mapping A named character vector where the names are names of
 #'   `var` columns in the new `AnnData` object and values are the names of
-#'   columns in the corresponding slot of `x`. See below for default if `NULL`
-#'   depending on the class of `x`.
+#'   columns in the corresponding slot of `x`. See below for details.
 #' @param obsm_mapping A named character vector where the names are keys of
 #'   `obsm` in the new `AnnData` object and values are the names of items in the
-#'   corresponding slot of `x`. See below for default if `NULL` depending on the
-#'   class of `x`.
+#'   corresponding slot of `x`. See below for details.
 #' @param varm_mapping A named character vector where the names are keys of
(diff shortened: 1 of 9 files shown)


R/as_AnnData.R
```diff
@@ -107,14 +107,35 @@
...
+#'   ## Graph conversion
+#'
+#'   By default, all graphs in a [`SeuratObject::Seurat`] object that match the
+#'   assay being converted are copied to the `obsp` slot of the new `AnnData`
+#'   object. If a graph does not have an associated assay:
+#'
+#'   - If `assay_name` is the default assay, they will be _converted_ with a
+#'     warning
+#'   - if `assay_name` is not the default assay, they will be _skipped_ with a
+#'     warning
```
lazappi, line 119: This logic might be a bit complex, we could make it always do one of these

R/from_Seurat.R
```diff
@@ -460,8 +499,26 @@ from_Seurat <- function(
...
+    graph_assay <- SeuratObject::DefaultAssay(graph)
+
+    # Handle graphs with missing assay information
+    if (is.null(graph_assay)) {
+      # If we are using the default assay, warn and convert
+      if (assay_name == SeuratObject::DefaultAssay(seurat_obj)) {
+        cli_warn(c(
+          "Graph {.val {graph_name}} does not have an associated assay",
+          "i" = "Assuming it belongs to the selected default assay ({.val {assay_name}})"
+        ))
```
rcannood, line 511: If we are in this edge case, will it still check whether the dimensions are correct?

lazappi: Yes, because the check for that is at the conversion step. So it will still get skipped if the dimensions don't match.

rcannood (approved):
LTGM!

PR #375, opened 2025-10-31
Add Bioconductor badges

lazappi:
## Description

Add Bioconductor badges to README

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # {anndataR}: An R package for working with AnnData objects <img src="man/figures/logo.png" align="right" alt="anndataR logo" width=120 />
 
 <!-- badges: start -->
-[![Lifecycle: experimental](https://img.shields.io/badge/lifecycle-experimental-orange.svg)](https://lifecycle.r-lib.org/articles/stages.html#experimental)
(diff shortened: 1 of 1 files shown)


lazappi:
Maybe we also want to update the lifecycle badge from "Experimental"?

rcannood (commented):
Looks great!

Let's indeed bump the lifecycle.

Do we need all of these badges? Some of these are definitely useful but others introduce more noise than signal at this stage. Or can we make a distinction between badges for users vs badges for developers?

lazappi:
> Let's indeed bump the lifecycle.

Done

> Do we need all of these badges? Some of these are definitely useful but others introduce more noise than signal at this stage. Or can we make a distinction between badges for users vs badges for developers?

Probably not, the function I used just added all of them. It's the same ones that are on the Bioconductor page if you want to pick some https://bioconductor.org/packages/release/bioc/html/anndataR.html.

rcannood (approved):

PR #392, opened 2025-12-03
Separate reading/writing keys in `HDF5AnnData`

### LINKED ISSUE #334: Avoid reading/writing all of `obs`/`var` when only names are needed
Currently for `HDF5AnnData`, `adata$obs_names()` requires reading/writing all of `obs` (same for `$var_names()`). We could make it so that only the names are read/written when that is all that is needed.

lazappi:
**Related to:** Fixes #334 

## Description

Allow reading/writing of keys in `HDF5AnnData` separately from the whole element (i.e. read the names of a list without reading the whole list):

- Add helper functions to read keys from H5AD files
- Add helper functions to write a data frame index to a H5AD file
- Use new helpers in `HDF5AnnData`
- Add new `*_keys()` methods to `HDF5AnnData`

## Checklist

**Before review**

- [x] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version


R/HDF5AnnData.R
```diff
@@ -433,6 +443,39 @@ HDF5AnnData <- R6::R6Class(
       self
     },
 
+    #' @description See [AnnData-usage]
+    obs_keys = function() {
+      names(self$obs)
+    },
+    #' @description See [AnnData-usage]
+    var_keys = function() {
+      names(self$var)
```
rcannood: this will still read in the entire df to get the colnames of the data frame, right?

rcannood (commented):
Great work!

One minor question.

R/HDF5AnnData.R
lazappi: Yes. We should use `read_h5ad_element_keys(..., dim = "cols')` here.

rcannood (approved):
LGTM, but the CI needs to be fixed.

rcannood:
Issues seem unrelated to this PR, so I'll merge it

PR #394, opened 2025-12-04
Force loading the environment in Python vignette

### LINKED ISSUE #386: scanpy and mudata are not available during pkgdown
On the website I see:

Image

↓

the code block that prints this message should get `echo=FALSE`. Also, we should make sure scanpy and mudata are available so the code chunks are evaluated when running pkgdown.

lazappi:
**Related to:** Fixes #386 
## Description

Add a `reticulate::py_config()` call to the Python usage vignette to force loading the Python environment

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/vignettes/usage_python.Rmd
+++ b/vignettes/usage_python.Rmd
@@ -12,6 +12,9 @@ vignette: >
 reticulate::py_require("mudata")
 reticulate::py_require("scanpy")
 
+# Force loading of the Python environment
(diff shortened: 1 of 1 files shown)


LouiseDck (approved):

rcannood (approved):
Thanks!

PR #403, opened 2025-12-21
Bump GHA cache version to force reinstall

Bisaloo:
**Related to:** https://github.com/scverse/anndataR/pull/190#issuecomment-3679016660

## Description

Force reinstall of Rhdf5lib / rhdf5 by creating a new package cache for the lint and pkgdown GitHub Action workflows. The same result can be achieved by manually deleting the existing caches rather than bumping the cache version.

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -30,6 +30,7 @@ jobs:
         with:
           extra-packages: any::lintr, any::styler, local::.
           needs: lint
+          cache-version: 2
 
       - name: Lint
         run: lintr::lint_package()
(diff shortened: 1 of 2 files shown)


lazappi:
Thanks for the PR! I've tried deleting the cache manually before but it's only a temporary fix. I think the issue might be that different actions are using the same cache name and there is a conflict. I've been meaning to check the names are unique but I haven't got to it yet. If you can check and update the PR that would be great.

Bisaloo:
Yes, you're probably right. The way to fix it should be to set a different cache version, which is then used to create the cache name, for the workflows generating conflicting caches, which is what this PR is doing.

lazappi:
Ah, right. That makes sense. Let's merge this and see if it helps.

lazappi (approved):

Issue #329, opened 2025-08-30
Unusually high memory usage for `{anndataR}` compared to Python `anndata`


yh464:
Dear anndataR team,

The memory usage for anndataR is much higher compared to the python interface of anndata. 
For the same file (https://datasets.cellxgene.cziscience.com/df829d08-839f-4f77-8581-5edc9705a5a7.h5ad), the memory usage from python is around 3.19 GB:

```
              total        used        free      shared  buff/cache   available
Mem:       65420264    26791720    33263416       30476     5365128    38050976
Swap:             0           0           0
```

And the memory usage from anndataR is around 5.95 GB, even when the data matrix is not loaded:
```
Error reading element X of type <csr_matrix>
'R_Calloc' could not allocate memory (1971718190 of 8 bytes)
> system('free')
              total        used        free      shared  buff/cache   available
Mem:       65420252    49919548    14864860       34552      635844    14910220
Swap:             0           0           0
```
After forcing garbage collection, memory is rapidly freed up:
```
> gc()
           used  (Mb) gc trigger    (Mb)   max used    (Mb)
Ncells  6336582 338.5   13590895   725.9    8839828   472.1
Vcells 13390530 102.2 4377879324 33400.6 5930311734 45244.7
> system('free')
              total        used        free      shared  buff/cache   available
Mem:       65420252     3617396    61166988       34552      635868    61212372
Swap:             0           0           0
```
The same out-of-memory error is raised even when trying to read the same dataset on a cluster with much larger RAM (45.44 GB), where the process is automatically killed.

Would you please look into this issue as there might be something in the backend that causes this excessive RAM usage? 

Many thanks,
YH

lazappi:
Could you please share the code you used to run this comparison? We haven't really optimised for memory usage so far so it's likely there are things we could do to improve this but there are also inherent difference between the two languages.

yh464:
Python code (3.13.3):
```
import scanpy as sc
adata = sc.read_h5ad('test.h5ad')
```

R code (R 4.5.1):
```
library(anndataR)
adata = read_h5ad('test.h5ad')
```

Simply that, nothing else - many thanks.

lazappi:
I've started to have a look into this. First thing is that the file is too big to load into R on my laptop (it work in Python though, so that's the fist sign), so I created a reasonably large subset (100,000 cells x 20,000 genes). I then created scripts with your example code and checked the memory usage:

```
[...]

PR #417, opened 2026-02-21
Optimise sparse matrix reading performance by avoiding `Matrix::sparseMatrix`

rcannood:
### Summary

Replace `Matrix::sparseMatrix()` with direct S4 construction (`new("dgCMatrix", ...)` / `new("dgRMatrix", ...)`) when reading sparse arrays from H5AD files. This eliminates the expensive COO→CSC/CSR coercion that `sparseMatrix()` performs internally, yielding **1.6–1.8× speedup** on sparse matrix reads.

### Problem

Profiling showed that `Matrix::sparseMatrix()` dominated read time for sparse H5AD files.
### Benchmark results

Dataset: 10,000 obs × 5,000 vars, generated with `dummy-anndata`, median of 5 runs.

#### `read_h5ad(as="InMemoryAnnData")`

| X type | Before | After | Speedup |
|---|---:|---:|---:|
| `float_matrix` (dense) | 0.853s | 0.789s | 1.08× |
| `float_csparse` (CSC) | 3.362s | 1.833s | **1.83×** |
| `float_rsparse` (CSR) | 3.098s | 1.825s | **1.70×** |
| `integer_matrix` (dense) | 1.067s | 1.057s | 1.01× |
| `integer_csparse` (CSC) | 3.277s | 2.107s | **1.56×** |
| `integer_rsparse` (CSR) | 3.296s | 2.106s | **1.57×** |

#### `hdf5_ad$X` (isolated slot read)

| X type | Before | After | Speedup |
|---|---:|---:|---:|
| `float_matrix` (dense) | 0.596s | 0.602s | 1.00× |
| `float_csparse` (CSC) | 2.779s | 1.650s | **1.68×** |
| `float_rsparse` (CSR) | 2.793s | 1.651s | **1.69×** |
| `integer_matrix` (dense) | 0.865s | 0.886s | 0.98× |
| `integer_csparse` (CSC) | 3.056s | 1.920s | **1.59×** |
| `integer_rsparse` (CSR) | 3.076s | 1.928s | **1.60×** |

Dense matrix performance is unchanged (not affected by this change). The remaining time is dominated by raw HDF5 I/O (`as.vector`, `.Call`/rhdf5) which is outside our control.

[results_after.txt](https://github.com/user-attachments/files/25460392/results_after.txt)  
[results_before.txt](https://github.com/user-attachments/files/25460393/results_before.txt)

## Checklist

**Before review**

- [x] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [x] Update `NEWS`
- [ ] Bump devel version


R/read_h5ad_helpers.R
```diff
@@ -266,29 +266,39 @@ read_h5ad_sparse_array <- function(
...
+  }
+  indices <- as.integer(as.vector(h5group$indices))
+  indptr <- as.integer(as.vector(h5group$indptr))
+  shape <- as.integer(as.vector(attrs[["shape"]]))
+
+  # The Matrix package validity checks require that indices are sorted within
+  # each major axis group (row indices within columns for CSC, column indices
+  # within rows for CSR). For sparse matrices in Python order isn't guaranteed,
+  # so we sort if needed. The vectorized approach uses order(group_id, index)
+  # with radix sort, which is fast even for large matrices.
```
LouiseDck: Is there a specific reason to explicitly go for radix sort? The default uses radix for short vectors and switches over to shell for large ones.

LouiseDck (approved):

lazappi:
Do we know if this has any effect on memory usage (as in #329)?

rcannood:
> Do we know if this has any effect on memory usage (as in #329)?

It does! I'm still learning to interpret the [report in bencher](https://bencher.dev/perf/anndatar/reports/b6c7058d-7871-491b-81db-7177b2eab32f?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=anndatar), but it looks like matrix-based operations have ~ -20% memory usage in this PR. 

> Is there a specific reason to explicitly go for radix sort? The default uses radix for short vectors and switches over to shell for large ones.

Because apparently radix should still outperform shell sort for large integer index vectors. However, I can probably try this out by removing it and seeing how the performance changes.

rcannood:
> > Is there a specific reason to explicitly go for radix sort? The default uses radix for short vectors and switches over to shell for large ones.
> 
> Because apparently radix should still outperform shell sort for large integer index vectors. However, I can probably try this out by removing it and seeing how the performance changes.

It seems that by changing the sort to auto, the memory usage in the test benches goes down:

image

( I think we need to compare the last two timepoints)

And execution time goes down in most cases, but goes up in others:

image

---

Therefore I'd propose we follow Louise's advice and just stick to using the default R order behaviour.

PR #418, opened 2026-02-21
Fix CI

rcannood:
Before we merge more PRs, it'd be nice if the CI was fixed :sweat_smile: 

## Checklist

**Before review**

- [ ] Add/update tests
- [x] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version


.github/workflows/R-CMD-check-bioc.yaml
```diff
@@ -47,69 +48,47 @@ jobs:
...
-          check_dir <- if (.Platform$OS.type == "windows") {
-            paste0(pkg_dir, ".Rcheck")
-          } else {
-            paste0(normalizePath(pkg_dir), ".Rcheck")
-          }
-          check_dir_norm <- normalizePath(check_dir)
-          cat("Check dir:", check_dir_norm, "\n")
-          cat("check-dir=", check_dir_norm, "\n", file = Sys.getenv("GITHUB_OUTPUT"), sep = "", append = TRUE)
-        shell: Rscript {0}
-
```
lazappi, line 88: I can't remember what all this was for, I think maybe it had something to do with exposing the output artifacts?

vignettes/usage_python.Rmd
```diff
@@ -9,15 +9,19 @@ vignette: >
...
+    reticulate::py_require("mudata")
+    reticulate::py_require("scanpy")
+    reticulate::py_config()
+    reticulate::py_module_available("scanpy") &&
+      reticulate::py_module_available("mudata")
+  },
+  error = function(e) {
+    FALSE
+  }
+)
```
lazappi: I agree we need this so checks will pass but we should be aware the code isn't being run and the vignette will be incomplete.

lazappi (approved):
Thanks for doing this! I've been meaning to get to it. It's annoying that `uv` isn't resolving the environment properly but I think it's something on their side, I checked and confirmed it's not related to **{anndataR}** or **{reticulate}**.

.github/workflows/R-CMD-check-bioc.yaml
rcannood: There must've been something wrong with the outputs of the bioc check step. Let me verify that it is currently working :)

rcannood: It is! Must have been an issue that's been solved in the meantime.

vignettes/usage_python.Rmd
rcannood: Fair enough, I suppose it's a bit of a shortcut to let the CI not test this at all.

But man, it was tough getting it to work Y_Y

lazappi (approved):

PR #446, opened 2026-04-15
Add Zarr to benchmarks

lazappi:
## Description

Add the Zarr backend to the benchmark tests

Needs #190 

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version


rcannood:
@lazappi Aside from the linting issue, is this PR ready for review?

lazappi:
I think so, now that the main PR is merged

Bisaloo:
Out of curiosity, would I be able to observe the impact of some changes in Rarr here?

I have my own continuous benchmarking suite but anndata(R) uses zarr in a different way which made me discover more performance issues last week. So I'd be interested in keeping an eye on your benchmarks.

rcannood:
> Out of curiosity, would I be able to observe the impact of some changes in Rarr here?

Hmm.. This might not be straight forward since you'll only know the impact of the changes you made in Rarr a new release was created.

I see Rarr has touchtone set  -- does this not do something similar?

Bisaloo:
> I see Rarr has touchtone set -- does this not do something similar?

It does. But when I worked with some anndata datasets last month, I discovered some inefficiencies I didn't suspect. In particular, because my current set of benchmarks only tests single arrays, instead of a collection of arrays as in anndata.

But based on your answer, the simplest solution is maybe to add an anndata dataset to my benchmarks then.

rcannood (approved):
LGTM!

PR #452, opened 2026-04-29
Write Zarr strings as VLen-UTF8

Bisaloo:
**Related to:** 

## Description

- [This follows the anndata file format spec](https://anndata.readthedocs.io/en/stable/fileformat-prose.html#string-arrays)
- This allows compatibility with zarr python for zarr version 3. In particular, this is blocking the addition of writing as zarr v3 because roundtrip with python fail.

## Checklist

**Before review**

- ~~Update and regenerate man pages~~
- [ ] Add/update tests
- ~~Add/update examples in vignettes~~
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/R/write_zarr_helpers.R
+++ b/R/write_zarr_helpers.R
@@ -412,9 +412,16 @@ write_zarr_string_array <- function(
     zarr_json_path <- file.path(store, name, "zarr.json")
     zarr_json <- jsonlite::read_json(zarr_json_path)
     zarr_json$data_type <- "string"
-    zarr_json$codecs <- c(
+    # There should be only one bytes-array codec
+    zarr_json$codecs <- lapply(
       zarr_json$codecs,
-      list(list(name = "vlen-utf8", configuration = list()))
+      function(codec) {
+         if (codec$name == "bytes") {
+           list(name = "vlen-utf8")
+         } else {
+           codec
+         }
+      }
     )
     jsonlite::write_json(
       zarr_json, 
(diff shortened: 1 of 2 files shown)


R/write_zarr_helpers.R
```diff
@@ -378,30 +378,66 @@ write_zarr_string_array <- function(
...
+      }
+    )
+    jsonlite::write_json(
+      zarr_json, 
+      zarr_json_path, 
+      auto_unbox = TRUE,
+      pretty = TRUE,
+      null = "null"
+    )
+  }
```
Bisaloo, line 433: This is not strictly need for now because writing is only possible for zarr v2, but this is laying the groundwork for an incoming PR from @Artur-man.

lazappi (commented):
I think this looks good. I'm just wondering if we should wait for a fix upstream rather than add a workaround here?

If we do it here, maybe some of the repeated code could be moved to a helper function?

Bisaloo:
> I'm just wondering if we should wait for a fix upstream rather than add a workaround here?

In theory, yes, I agree.

The reality is that I'm spread quite thin over a large number of projects and extending the writing capabilities of Rarr (as opposed to reading) is somewhat lower priority at the moment.

I could submit a follow up PR once the new interface is in place (I'd love to get to it before the next release but I cannot say for sure).

Artur-man:
You guys let me know how you wanna move forward. 

I can also write on top these commits of Hugo and open the zarr v3 write PR immediately, thus you do not need to work on it much @Bisaloo.

Artur-man:
I will open a PR on top of @Bisaloo's commits, lets get this rolling ....

lazappi:
Is all this covered by #455 or will this need to be merged first?

Artur-man:
> Is all this covered by https://github.com/scverse/anndataR/pull/455 or will this need to be merged first?

all these changes are available in #455. But if you wanna merge this first, it is up to you.

rcannood:
#455 was merged!

PR #455, opened 2026-05-10
Write support for anndata-zarr v3

Artur-man:
**Related to:** #446 #452 #91 

## Description

This PR introduces write support for anndata objects in zarr v3 format. Integration (write, roundtrip etc.) tests are updated to test both zarr v2 and v3. Written by @Bisaloo and @Artur-man.

Similar to `scverse/anndata`, the default write format is controlled using `anndataR.zarr_format` and set by, for example:

```
options(anndataR.zarr_format = 3)
``` 

There are a few number of issues needs to be resolved in `Rarr` package to pass all the tests, which we will tackle during the PR. 

Supersedes #452 

## Checklist

**TODO**

- [x] Push final changes after Rarr 2.1.13 is in Bioc devel
- [x] Change `zarr_version` to `zarr_format` for clarity 

**Before review**

- [x] Add/update tests
- [x] Pass CI checks

**Before merge**

- [x] Update `NEWS`
- [ ] Bump devel version


Artur-man:
Thanks to @Bisaloo all my tests are passing on my local. I will push the changes once bioc devel is updated.

Artur-man:
@lazappi @rcannood @LouiseDck 

The PR is ready for review now. I have also updated vignettes to some degree but let me know if I missed anything. 

Only coverage and ubuntu-release are failing but this looks like because of BioC-devel updates of Rarr.

R/write_zarr.R
```diff
@@ -16,7 +16,8 @@
 #'   * `r+` opens an existing file for read/write
 #'   * `w` creates a file, truncating any existing ones
 #'   * `w-`/`x` are synonyms creating a file and failing if it already exists
-#' @param ... Additional arguments passed to [as_AnnData()]
+#'
+#' @param ... Additional arguments passed to [as_AnnData()] or [as_ZarrAnnData()]
```
lazappi, line 20: I think what was there previously is fine

R/write_zarr_helpers.R
```diff
@@ -371,41 +389,83 @@ write_zarr_string_array <- function(
...
   name,
   compression,
+  zarr_format = .get_zarr_format(),
   version = "0.2.0"
 ) {
   dims <- dim(value) %||% length(value)
 
+  # if dims is zero, fix chunk dim to 1, but raises warnings
+  # https://github.com/Huber-group-EMBL/Rarr/issues/89
+  chunk_dims <- vapply(dims, \(.) if (.) . else 1, numeric(1))
```
lazappi: I think something like this would be clearer

```suggestion
  chunk_dims <- ifelse(dims == 0, 1, dims)
```

R/write_zarr_helpers.R
```diff
@@ -371,41 +389,83 @@ write_zarr_string_array <- function(
...
+  # suppress chunk dim warnings
+  suppressWarnings({
     Rarr::create_empty_zarr_array(
       file.path(store, name),
       dim = dims,
-      chunk_dim = dims,
-      data_type = "<U",
-      nchar = 1,
+      chunk_dim = chunk_dims,
+      order = if (length(dims)) "C" else "F",
```
lazappi: Can you explicitly test against `NULL` or 0 (whatever the case is)?

R/ZarrAnnData.R
```diff
@@ -511,6 +525,8 @@ ZarrAnnData <- R6::R6Class(
 #'   * `r+` opens an existing file for read/write
 #'   * `w` creates a file, truncating any existing ones
 #'   * `w-`/`x` are synonyms, creating a file and failing if it already exists
+#' @param zarr_format The Zarr data format to be used for [write_zarr()]:
+#'   "2" for Zarr v2, and "3" for Zarr v3
```
lazappi: Should the values be strings, integers or numeric? This looks like they should be strings but previously I think they were integers.

tests/testthat/test-roundtrip-empty.R
```diff
@@ -10,13 +10,22 @@ tryCatch(ad$settings$allow_write_nullable_strings <- TRUE, error = function(e) {
...
 known_issues <- read_known_issues()
 
 name <- "empty"
 
-for (fmt in c("h5ad", "zarr")) {
+for (fmt in c("h5ad", "zarrv2", "zarrv3")) {
   fmt_config <- get_fmt_config(fmt)
 
+  if (grepl("zarr", fmt_config$ext, fixed = TRUE)) {
+    options(anndataR.zarr_format = fmt_config$zarr_format)
```
lazappi: It might be better to use `withr::with_options()` to set this temporarily.

vignettes/anndataR.Rmd
```diff
@@ -223,6 +225,23 @@ tmpfile <- tempfile(fileext = ".zarr")
 write_zarr(obj, tmpfile)
 ```
 
+Two most recent Zarr formats are supported: Zarr v2 and v3. We can switch between two when writing objects to `.zarr` stores.
```
lazappi:
```suggestion
We support reading and writing Zarr v2 and v3.
Zarr v3 is the default, to use Zarr v2 globally set `options(anndataR.zarr_format = 2)` or use `zarr_format = 2` to set it in an individual function all. 
```

vignettes/anndataR.Rmd
```diff
@@ -213,6 +213,8 @@ write_h5ad(obj, tmpfile)
 Similarly, we can write `AnnData` and other objects to `.zarr` stores too.
 
 ```{r write-to-disk-zarr}
+options(anndataR.zarr_format = 2)
```
lazappi: Does this need to be set here?

DESCRIPTION
```diff
@@ -53,7 +53,7 @@ Suggests:
     knitr,
     processx,
     rhdf5 (>= 2.52.1),
-    Rarr (>= 1.11.12),
+    Rarr,
```
lazappi: Do we need to bump the minimum version here? I don't think we should remove it entirely.

R/Rarr_utils.R
```diff
@@ -103,3 +107,9 @@ zarr_path_exists <- function(store, target_path) {
     }
   }
 }
+
+#' @noRd
+.get_zarr_format <- function() {
+  vr <- getOption("anndataR.zarr_format")
+  if (is.null(vr)) 2 else vr
+}
```
lazappi: I'm not sure we need this helper, we can just use `getOption("anndataR.zarr_format", 2L)`. I would only keep it if it is going to some extra validation (which is maybe not a bad idea).

If we do keep it, there should be a matching setter and they should be exported, I don't like seeing `zarr_format = .get_zarr_format()` in exported function arguments when the getter isn't accessible to users.

lazappi (changes requested):
There are a few fairly minor things I would like to clean up

R/Rarr_utils.R
Artur-man: makes sense

vignettes/anndataR.Rmd
Artur-man: I think this shows the user how to change the version, as an example

Artur-man: If we are setting version 3 as default, this needs to be shown how to set version to 2. Otherwise, we should do the opposite.

R/write_zarr_helpers.R
Artur-man: I do not remember why I distinguish "C" and "F" ordering ... should be to mimic anndata-py. Lets check that first ...

R/Rarr_utils.R
Artur-man: so only ?:

```
write_zarr_categorical <- function(
  value,
  store,
  name,
  compression,
  zarr_format = getOption("anndataR.zarr_format"),
  version = "0.2.0"
```

Artur-man: we can remove it from exported arguments if it looks bad ... but not using a getter will make things hard ... what if the value of getOption("anndataR.zarr_format") is not set ? ... then the package wont have a mech. of choosing default.

tests/testthat/test-roundtrip-empty.R
Artur-man: Hmm how are we going to do it conditionally though ?

Artur-man: I should also wrap the if statement with withr too ?

Artur-man:
```
  withr::with_options({
    if (grepl("zarr", fmt_config$ext, fixed = TRUE)) {
      ad$settings$zarr_write_format <- fmt_config$zarr_format
      options(anndataR.zarr_format = fmt_config$zarr_format)
    }
  }, {
    for (name in test_names) {
      # first generate a python adata
      adata_py <- da$generate_dataset(
        x_type = name,
        obs_types = list(),
...
```

R/write_zarr_helpers.R
```diff
@@ -130,10 +132,10 @@ write_zarr_element <- function(
 #' @param name Name of the element within the Zarr store
 #' @param encoding The encoding type to set
 #' @param version The encoding version to set
-write_zarr_encoding <- function(store, name, encoding, version) {
+write_zarr_encoding <- function(store, name, encoding, version, zarr_format) {
   Rarr::write_zarr_attributes(
     file.path(store, name),
-    new.zattrs = list(`encoding-type` = encoding, `encoding-version` = version)
+    new.zattrs = list(`encoding-type` = encoding, `encoding-version` = version),
```
Bisaloo, line 141:
```suggestion
    new.zattrs = list(`encoding-type` = encoding, `encoding-version` = version),
    zarr_version = zarr_format
```
?

DESCRIPTION
Bisaloo: https://github.com/Huber-group-EMBL/Rarr/pull/181


```suggestion
    Rarr (>= 2.1.13),
```

tests/testthat/test-roundtrip-empty.R
Artur-man: @lazappi should you check if `withr` is installed in these tests, I do not see any checks in other tests, e.g. `test-HDF5AnnData.R`

DESCRIPTION
Artur-man: done @Bisaloo

R/write_zarr_helpers.R
Artur-man: @Bisaloo Hmm, does Rarr automatically detect if the zarr store we write attr to is either 2 or 3 ? otherwise I am ok.

vignettes/anndataR.Rmd
Artur-man: We should first decide what should be the default. Better to go with anndata's default, @ilan-gold ?

ilan-gold: I would go with v3, yeah.

R/Rarr_utils.R
Artur-man: I am removing `.get_zarr_format` from oxygen documentations, but lets keep the function internal.

R/ZarrAnnData.R
Artur-man: done, should be integer

R/write_zarr_helpers.R
Bisaloo: It will look at the target folder to see if it already contains v2 or v3 metadata, but not at other elements in the store (parent group, etc.).
If I remember correctly, in your case, you always start by initializing the group, so the version should be auto-detected, but the trailing comma should be removed in this case.

Artur-man: ah ok but I have already pushed a version where the zarr format is specified ... lets move on its no big deal!

R/write_zarr_helpers.R
Artur-man: so the value is only passed to write_string_<type>() only if its an array/vector of characters, what do you want to be tested exactly ?

R/Rarr_utils.R
lazappi: If you use `getOption("anndataR.zarr_format", 2L)` the second value is what gets returned if the option is not set. You can also do something like `zarr_format = NULL` and be clear in the documentation that it reads from the argument and/or has some default fallback.

Artur-man: `getOption("anndataR.zarr_format", 2L)` makes sense.

Artur-man: I have put `getOption("anndataR.zarr_format", 2L)` directly in the argument, I think it looks clean.

Artur-man:
there is a problem with a main branch commit I merged earlier, failing tests ... checking now.

R/write_zarr_helpers.R
lazappi: Instead of `if (length(dims))` use `if (length(dims) == 0/NULL)` so it's explicit what is being tested against.

tests/testthat/test-roundtrip-empty.R
lazappi: It's a suggested dependency so it should be available during testing, it's used in the roundtrip tests. I guess it we want to use it like this we should move it to `imports`.

Artur-man: Hmm, if the package is only used in tests, why do we add it to imports ... I would rather 

1. throw an `skip_if_not_installed("withr")` on top, like you have done with `skip_if_no_dummy_anndata()`, or
2. use `rlang::check_installed` or `rlang::is_installed` to check `withr`, since **rlang** is already imported.

R/write_zarr_helpers.R
Artur-man: I think `if (length(dims) > 0)` or `if (length(dims) != 0)` would suffice then.

R/ZarrAnnData.R
rcannood: this should receive the format from `private$.zarrformat`, right?

R/Rarr_utils.R
```diff
@@ -7,12 +7,16 @@ ZARR_METADATA_FILES <- c(".zarray", ".zattrs", ".zgroup", "zarr.json")
...
+#' @param format Zarr format
 #'
 #' @return `NULL`
 #'
 #' @noRd
-create_zarr_group <- function(store, name, version = "v2") {
+create_zarr_group <- function(
+  store,
+  name,
+  format = getOption("anndataR.zarr_format", 3L)
```
rcannood: I feel like helper functions probably shouldn't have a default value here, so that if it is ever accidentally not passed correctly, we don't start creating mixed version zarr files

R/Rarr_utils.R
```diff
@@ -51,15 +59,15 @@ create_zarr_group <- function(store, name, version = "v2") {
...
 #'
 #' @param store The location of the Zarr store
-#' @param version Zarr version
+#' @param format Zarr format
 #'
 #' @return `NULL`
 #'
 #' @noRd
-create_zarr <- function(store, version = "v2") {
+create_zarr <- function(store, format = getOption("anndataR.zarr_format", 3L)) {
```
rcannood: same here; I feel like helper functions probably shouldn't have a default value here, so that if it is ever accidentally not passed correctly, we don't start creating mixed version zarr files

R/ZarrAnnData.R
rcannood: pass format

R/write_zarr_helpers.R
```diff
@@ -419,12 +524,18 @@ write_zarr_string_array <- function(
...
 #' @param version Encoding version of the element to write
 write_zarr_categorical <- function(
   value,
   store,
   name,
   compression,
+  zarr_format = getOption("anndataR.zarr_format", 3L),
   version = "0.2.0"
 ) {
   create_zarr_group(store, name)
```
rcannood: need to pass zarr format?

R/write_zarr_helpers.R
```diff
@@ -472,26 +592,79 @@ write_zarr_categorical <- function(
...
     compressor = .get_compressor(compression),
-    zarr_version = 2L
+    zarr_version = zarr_format
+  )
+  # Rarr doesn't yet provide an explicit interface to deal with VLen-UTF8 so
+  # we patch the metadata by hand.
+  # Will be resolved by https://github.com/Huber-group-EMBL/Rarr/issues/111.
+  if (zarr_format == 2L) {
+    zarr_json_path <- file.path(store, name, ".zarray")
+    zarr_json <- jsonlite::read_json(zarr_json_path)
```
rcannood: jsonlite is not declared as a dependency

R/write_zarr.R
```diff
@@ -44,6 +45,15 @@
...
 #' adata$write_zarr(zarr_store)
 #'
+#' # set Zarr version to 2.
+#' options(anndataR.zarr_format = 2)
+#' zarr_store <- tempfile(fileext = ".zarr")
+#' adata$write_zarr(zarr_store)
+#'
+#' # write .zarr as version 3
+#' zarr_store <- tempfile(fileext = ".zarr")
+#' adata$write_zarr(zarr_store, zarr_format = 3)
```
rcannood: cam we add a unit test that tests whether we can write and read files with v2 and v3, and see if we can read these out again in python?

R/write_zarr_helpers.R
```diff
@@ -7,6 +7,11 @@
 #' @param name Name of the element within the Zarr store
 #' @param compression The compression to use when writing the element. Can be
 #' one of `"none"` or `"gzip"`. Defaults to `"none"`.
+#' @param zarr_format The format to use when writing the Zarr file.
+#'   Should be either 2 or 3 for Zarr v2 or v3 formats, respectively.
+#'   Unless it is specified, Zarr v3 will be used by default.
+#'.  The format can also be specified using
+#'   `anndataR.zarr_format`, e.g. `options(anndataR.zarr_format = 2)`.
```
rcannood: this typo (`#'.`) is copy pasted x number of times

R/write_zarr_helpers.R
rcannood: i agree, please fix

R/ZarrAnnData.R
```diff
@@ -338,6 +351,7 @@ ZarrAnnData <- R6::R6Class(
 
       private$.compression <- compression
 
+      private$.zarrformat <- zarr_format
```
rcannood: when we're reading an existing file, shouldn't we compare the current on-disk zarr_format to the argument passed in this function? because if the current ondisk format is 2, but we store 3 because that's what gets passed via `zarr_format = getOption(...)`, we might end up with a mixed store.

tests/testthat/test-Zarr-read.R
```diff
@@ -118,8 +114,8 @@ for (zarr_version in c("v2", "v3")) {
     expect_equal(dim(array), c(5, 10))
   })
 
-  test_that(paste("reading Zarr", zarr_version, "mappings works"), {
-    if (zarr_version == "v3") {
+  test_that(paste("reading Zarr", zarr_format, "mappings works"), {
+    if (zarr_format == "3") {
       # TODO: Remove when v3 recarray support is implemented
```
rcannood, line 119: this should be recorded in `inst/known_issues.yaml` and be tested against

rcannood (changes requested):
There seem to be multiple issues in this PR. Overall, I would not add default values on internal functions like `zarr_format = getOption("anndataR.zarr_format", 3L)`, because this makes the code prone to accidentally forgetting to pass the zarr_format. Instead, define these arguments in the very few entrypoints used by the end user and propagate from there.

With the current implementation, it's quite easy to get mixed format zarr stores:

```r
# remotes::install_github("Huber-group-EMBL/Rarr")
devtools::load_all()

ad <- AnnData(
  X   = matrix(1:6, 2, 3),
  obs = data.frame(row.names = c("c1", "c2"), grp = factor(c("a", "b"))),
  var = data.frame(row.names = c("g1", "g2", "g3"), gene = 1:3)
)

# set the default zarr format to 3
options(anndataR.zarr_format = 3)

# write to v2 zarr format (should override the default option)
store <- tempfile(fileext = ".zarr")
write_zarr(ad, store, zarr_format = 2)
# Warning message:
# Could not write element 'obs/grp' of type 'factor': The path contains both `.zattrs` (v2 Zarr specification) and `zarr.json` (v3 Zarr specification) files, which is not allowed. 

# Root node: written as v3 (option default), NOT v2 as requested  -> A1
file.exists(file.path(store, "zarr.json")) # TRUE (v3)
file.exists(file.path(store, ".zgroup")) # FALSE

# X array: written as v2 (uses the argument)
file.exists(file.path(store, "X", ".zarray")) # TRUE (v2)

# Categorical group obs/grp: v3 wrapper (option default) around v2 children -> A3
file.exists(file.path(store, "obs", "grp", "zarr.json"))  # TRUE  (v3 group)
file.exists(file.path(store, "obs", "grp", "codes", ".zarray"))  # TRUE (v2 child)
```

---

we need to add this as a regression test to verify that setting one option but explicitly asking for another version works.

R/Rarr_utils.R
Artur-man: I agree, thanks for the catch ... will fix em

R/write_zarr_helpers.R
Artur-man: hate when this happens ...

R/write_zarr.R
Artur-man: I thought this happens in round trip tests, no ?

R/write_zarr_helpers.R
Bisaloo: One technical solution is to use roxygen's `@inheritParams`. It's used in one other place in the codebase already

Artur-man: There are also lots replications in roxygen2 documentation across the package (not just zarr). Perhaps we should open an issue/PR for that too.

tests/testthat/test-Zarr-read.R
Bisaloo: Was this a limitation in Rarr? If so, it should be solved since https://github.com/Huber-group-EMBL/Rarr/pull/174

tests/testthat/test-roundtrip-empty.R
Artur-man: what should we do here ? shall I commit or shall we fix this in another PR, checking `withr` should be addressed everywhere in the package not just in zarr scripts/tests.

R/write_zarr_helpers.R
Artur-man: hmm weird, how is this passing the check ?

Artur-man: @rcannood should it go to suggests or imports ?

Artur-man:
> There seem to be multiple issues in this PR. Overall, I would not add default values on internal functions like zarr_format = getOption("anndataR.zarr_format", 3L), because this makes the code prone to accidentally forgetting to pass the zarr_format. Instead, define these arguments in the very few entrypoints used by the end user and propagate from there.

@rcannood, we only enter default arg values in the following: 

```
(base) LANDTHALER57:anndataR amanuky$ git grep 'getOption("anndataR.zarr_format", 3L)'
R/AbstractAnnData.R:      zarr_format = getOption("anndataR.zarr_format", 3L)
R/ZarrAnnData.R:      zarr_format = getOption("anndataR.zarr_format", 3L)
R/ZarrAnnData.R:  zarr_format = getOption("anndataR.zarr_format", 3L)
man/AbstractAnnData.Rd:  zarr_format = getOption("anndataR.zarr_format", 3L)
man/ZarrAnnData.Rd:  zarr_format = getOption("anndataR.zarr_format", 3L)
man/as_ZarrAnnData.Rd:  zarr_format = getOption("anndataR.zarr_format", 3L)
```

Artur-man:
@rcannood gave you access to the fork if you wanna merge directly, otherwise I will check soon thanks!

rcannood (changes requested):
I found a few more issues. @Artur-man can you review and merge the PRs? After they are merged I can approve and merge.

Artur-man:
Seems bioc is still down ...

PR #387, opened 2025-12-01
Add support for `DelayedArray` reading

lazappi:
**Related to:** Fixes #128 

## Description

Add support for backed objects using `DelayedArray` matrices:

- Add a `HDF5File` class to store file handles and paths with convenient methods for opening/closing them
- Use the **{HDF5Array}** package for reading dense and sparse arrays from HDF5 files
- Add a `backed` argument to `HDF5AnnData`
- Add the `backed` option to `InMemoryAnnData`
- Conversion of backed `AnnData` objects to `SingleCellExperiment`/`Seurat`

Other features that could be added:

- Add support for writing `DelayedArray` matrices
- Add `SparseArray` as a format for in-memory matrices

## Checklist

**Before review**

- [x] Update and regenerate man pages
- [x] Add/update tests
- [ ] Add/update examples in vignettes
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

PR #471, opened 2026-06-21
Delayed array improvements

rcannood:
**Related to:** 

## Description

This PR aims to solve several issues in #387 ;

- Partially fixes performance regression: eager reads using HDF5Array seems to be quite a bit slower than native rhdf5 reads (110 alerts). the proposed changes bring it down to 10 alerts.
- anndataview now correctly subsets DelayedArrays instead of keeping them as is.
- When converting a backed HDFAnnData to InMemoryAnnData, DeferredArrays are materialised
- DelayedArrays are materialised on write. I don't think there is a way around this in HDF5AnnData.
- Fixed DelayedMatrix getting transposed on SCE roundtrip (cfr. Louise's comment)

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/R/HDF5AnnData.R
+++ b/R/HDF5AnnData.R
@@ -558,6 +558,20 @@ HDF5AnnData <- R6::R6Class(
     #' @description Close the HDF5 file handle
     close = function() {
       private$.hdf5_file$close()
+    },
+
+    #' @description Convert to an [`InMemoryAnnData`]. Any backed
+    #'   (`DelayedArray`) slots are materialized into ordinary in-memory
+    #'   matrices: an in-memory object should not stay tied to an on-disk file.
+    as_InMemoryAnnData = function() {
+      if (isTRUE(private$.backed)) {
+        prev <- private$.backed
+        private$.backed <- FALSE
+        on.exit(private$.backed <- prev, add = TRUE)
+        # Hold the file open for the whole multi-slot read (single handle).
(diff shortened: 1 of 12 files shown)


rcannood:
@lazappi You were right; the previous PR was unnecessarily complicated. I split the most relevant changes up into separate commits so it's easier to review. I also got rid of the `with_handle()` but kept the rest of the changes. Would you be able to re-review? :bow:

Note, I didn't add a warning when a user tries to convert a backed hdf5anndata to inmemoryanndata. I feel like when they trigger this action, you sortof expect to run out of memory if you're doing this on a large anndata.

R/HDF5AnnData.R
```diff
@@ -393,6 +393,14 @@ HDF5AnnData <- R6::R6Class(
       }
       private$.backed <- backed
```
lazappi, line 394: Did you want to change this to `as_delayed` (or just `delayed`)?

lazappi (approved):
You were right, this was much easier to review 😸!

R/HDF5AnnData.R
rcannood: I do like as_delayed better, but at this stage I'm not that bothered by backed anymore :P

PR #478, opened 2026-07-11
Fix `create_zarr()` so that Zarr stores can be created at relative paths

Artur-man:
## Description

Small fix to `create_zarr` internal that safely creates a store using `dirname` instead of `gsub`, otherwise creating a store in the way directory, e.g. `create_zarr("temp.zarr")`, fails. 

## Checklist

**Before review**

- [x] Update and regenerate man pages (no need)
- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # anndata 1.3.0
 
 - Bioconductor 3.24 devel
+- Fix `create_zarr()` so that Zarr stores can be created at relative paths and at paths containing regex metacharacters (PR #478).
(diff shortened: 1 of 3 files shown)


rcannood (approved):
Good catch!

I've pushed two commits to your branch:

- Added a news entry
- Added regression test

This PR should also be backported to the RELEASE_3_23 branch, which I will do later.

PR #479, opened 2026-07-11
Simplify roxygen docs

Artur-man:
**Related to:** #455 and https://github.com/scverse/anndataR/pull/455#discussion_r3473328944

## Description

Based on @Bisaloo 's suggestion, using `@inheritParams` for h5ad/zarr helpers to avoid repetitive `@param` tags. For now, only applied to helper scripts, but could be performed elsewhere if needed. 

Another note is that, since these are internal functions and are not exported, `@inheritParams`, does not really help beyond telling devs where to look for argument descriptions, but helps avoiding errors. I have found many inconsistencies while replacing with `@inheritParams` tag.  

## Checklist

**Before review**

- [x] Add/update tests (no need)
- [x] Add/update examples in vignettes (no need)
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/R/read_h5ad_helpers.R
+++ b/R/read_h5ad_helpers.R
@@ -1,32 +1,3 @@
-#' Read H5AD encoding
-#'
-#' Read the encoding and version of an element in a H5AD file
-#'
-#' @param hdf5_file An `HDF5File` object
-#' @param name Name of the element within the H5AD file
-#'
-#' @return A named list with names type and version
-#'
-#' @noRd
-read_h5ad_encoding <- function(hdf5_file, name) {
-  hdf5_file$open_and_defer_close(readonly = TRUE)
-
-  tryCatch(
-    {
-      attrs <- rhdf5::h5readAttributes(hdf5_file$handle, name)
-      list(
-        type = attrs[["encoding-type"]],
-        version = attrs[["encoding-version"]]
-      )
-    },
-    error = function(e) {
-      cli_abort(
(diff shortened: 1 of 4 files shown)


R/read_h5ad_helpers.R
```diff
@@ -313,12 +300,7 @@ read_h5ad_csc_matrix <- function(hdf5_file, name, version, backed, ...) {
 #'
 #' Read a sparse array from an H5AD file
 #'
-#' @param hdf5_file An `HDF5File` object
-#' @param name Name of the element within the H5AD file
-#' @param backed Whether to return a DelayedArray
-#' @param version Encoding version of the element to read
-#' @param type Type of the sparse matrix, either "csr_matrix" or "csc_matrix"
```
rcannood, line 320: The value for `type` shouldn't be inherited here, because it provides information beyond what is in `read_h5ad_element`. Alternatively, they could be moved to the description of this function.

rcannood (changes requested):
Looks quite good to me. Are there any differences in docs beyond what I commented on?

Artur-man:
@rcannood I also added a missing `chunk_size` param at the top of helpers.

R/read_h5ad_helpers.R
Artur-man: fixed. Please let me know if you see something else to be fixed.

Issue #382, opened 2025-11-20
Explore alternative matrix classes


lazappi:
See what improvements can be gained by using alternative classes to store matrix data, mostly the Bioconductor [**{SparseArray}**](https://www.bioconductor.org/packages/release/bioc/html/SparseArray.html) package.

This should help with #301 and possibly #329.

PR #483, opened 2026-07-20
Avoid copy when dropping attributes

Bisaloo:
## Description

I may be missing something so mostly opening this for discussion for now.

In many cases, it seems anndataR wants a vector instead of an array with a single dimension. In the current code, this is usually achieved via `as.vector()` / `as.integer()` / etc.

But this is very inefficient because it results in a entire new copy of the object in memory.

Would it work for you to drop the `dim` attribute? In your use case, it should not result in an additional copy of the data.

``` r
x <- sample(1:1e7, 1e7, replace = TRUE)
dim(x) <- 1e7

bench::mark(
  as.vector(x),
  as.integer(x),
  # We assign to y here for benchmarking purposes. To avoid removing dim for our test object after the first iteration.
  { 
    y <- x
    dim(y) <- NULL
    y
  },
  iterations = 1000
)
#> # A tibble: 3 × 6
#>   expression                       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 as.vector(x)                  11.5ms   12.3ms      81.3    38.1MB     25.3
#> 2 as.integer(x)                 11.5ms   12.3ms      81.2    38.1MB     20.3
#> 3 { y <- x dim(y) <- NULL y }    326ns    359ns 2545301.         0B      0
```

<sup>Created on 2026-07-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I took a reasonably large test dataset but note that this get much worse with larger datasets, while remove `dim` always remains free, independent of the dataset size.

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

### DIFF
--- a/R/read_zarr_helpers.R
+++ b/R/read_zarr_helpers.R
@@ -47,8 +47,8 @@ read_zarr_element <- function(
     "numeric-scalar" = read_zarr_numeric_scalar,
     "categorical" = read_zarr_categorical,
     "string-array" = read_zarr_string_array,
-    "nullable-integer" = read_zarr_nullable_integer,
-    "nullable-boolean" = read_zarr_nullable_boolean,
+    "nullable-integer" = read_zarr_nullable,
+    "nullable-boolean" = read_zarr_nullable,
     cli_abort(
       "No function for reading Zarr encoding {.cls {type}} for element {.val {name}}"
     )
@@ -263,32 +263,6 @@ read_zarr_rec_array <- function(store, name, version = "0.2.0") {
   version <- match.arg(version)
   Rarr::read_zarr_array(file.path(store, name)) |>
     lapply(as.vector)
-}
-
-#' Read Zarr nullable boolean
-#'
-#' Read a nullable boolean from a Zarr store
-#'
-#' @inheritParams read_zarr_element
-#'
-#' @return A boolean vector
-#'
-#' @noRd
-read_zarr_nullable_boolean <- function(store, name, version = "0.1.0") {
-  as.logical(read_zarr_nullable(store, name, version))
-}
-
-#' Read Zarr nullable integer
-#'
-#' Read a nullable integer from a Zarr store
-#'
-#' @inheritParams read_zarr_element
-#'
-#' @return An integer vector
-#'
-#' @noRd
-read_zarr_nullable_integer <- function(store, name, version = "0.1.0") {
-  as.integer(read_zarr_nullable(store, name, version))
 }
 
 #' Read Zarr nullable
--- a/R/read_zarr_helpers.R
+++ b/R/read_zarr_helpers.R
@@ -431,8 +431,8 @@ read_zarr_data_frame <- function(
 #' @noRd
 read_zarr_collection <- function(store, name, item_names) {
   items <- lapply(
-    item_names,
-    function(item_name) {
+    file.path(name, item_names),
+    function(new_name) {
       new_name <- paste0(name, "/", item_name)
       encoding <- read_zarr_encoding(store, new_name)
       read_zarr_element(
--- a/R/read_zarr_helpers.R
+++ b/R/read_zarr_helpers.R
@@ -229,15 +229,18 @@ read_zarr_sparse_array <- function(
 
   attrs <- Rarr::read_zarr_attributes(file.path(store, name))
 
+  data <- Rarr::read_zarr_array(file.path(store, name, "data"))
+  dim(data) <- NULL
+  indices <- Rarr::read_zarr_array(file.path(store, name, "indices"))
+  dim(indices) <- NULL
+  indptr <- Rarr::read_zarr_array(file.path(store, name, "indptr"))
+  dim(indptr) <- NULL
+
   construct_sparse_matrix(
-    data = as.vector(Rarr::read_zarr_array(file.path(store, name, "data"))),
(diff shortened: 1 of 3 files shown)


rcannood:
Looks interesting! If it improves performance, I'd definitely be down! :)

rcannood:
LGTM when you feel this PR is ready for a review, or what additional changes you think need to be made :)

Bisaloo:
On a 413 MB "X" layer:

```r
cross::bench_branches(
  {
    bench::mark(
      anndataR:::read_zarr_sparse_array("adata", "X"),
      iterations = 50
    )
  },
  branches = "devel"
)
```

```
✔ Installing branch 'fewer-conversion'
✔ Installing branch 'devel'
✔ Running `expr` across variants
# A tibble: 2 × 14
Loading required namespace: Matrix
  branch           expression                                                min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result             memory     time       gc      
  <chr>            <bch:expr>                                            <bch:t> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>             <list>     <list>     <list>  
1 fewer-conversion "anndataR:::read_zarr_sparse_array(\"adata\", \"X\")"   7.13s  8.68s     0.114    17.9GB    0.629    50   276      7.32m <dgRMatrx[,19244]> <Rprofmem> <bench_tm> <tibble>
2 devel            "anndataR:::read_zarr_sparse_array(\"adata\", \"X\")"    8.3s  9.11s     0.108    17.9GB    0.558    50   258      7.71m <dgRMatrx[,19244]> <Rprofmem> <bench_tm> <tibble>
```

Not as huge as what I'd have hoped but I think it can still be noticeable, especially on spatialdata objects with multiple anndata components.

This is good to go as far as I'm concerned.

Bisaloo:
Related:

This step:

https://github.com/scverse/anndataR/blob/7b7737cc6ae06abfb8dbf1d4c2813c3eaabd1e32/R/utils.R#L297

is slow and somewhat unintuitive.

I have an X layer where the `data` component is nicely saved as int64. This `as.double()` triggers an additional, expensive, copy of the data.

But it seems that if we leave `data` as integer, the Matrix package emits warnings. This is thus related to https://github.com/scverse/anndataR/issues/382.

rcannood:
I got curious and tried to reproduce the numbers on my end. It'd indeed more subtle than we both hoped. I think this is because `dim<-` is only free when the object has refcount 1, and `Rarr::read_zarr_array()` binds its result to a local before returning it, so the value comes back already shared and `dim<-` duplicates exactly like `as.vector()` did.

On a 5M-nnz CSR store:

* total allocation of `read_zarr_sparse_array()` is byte-identical -- 1,271,462,968 on both branches
* Rprofmem logs the same 40,000,048-byte allocation in both: on devel as `"as.vector" "construct_sparse_matrix" ...`, on your branch attributed straight to `read_zarr_sparse_array`.

Merging `read_zarr_nullable_integer()` / `read_zarr_nullable_boolean()` into one does make sense, because Rarr already outputs the right vector data types. Could we spin this off into a separate pr?

Would it make sense to take a look into the Rarr::read_zarr_array implementation? Claude reports: `read_zarr_array()` accounted for ~1.09GB of the 1.19GB and ~950ms of the ~980ms in, so roughly 18x the 60MB payload. Almost all of it sits in `.chunk_positions_by_chunk`, in `split()` / `as.factor()` / `relist()` / `unique()`. [profile.zip](https://github.com/user-attachments/files/30457885/profile.zip)

Bisaloo:
@rcannood, before I attempt to reproduce, could you confirm you are using the latest version of Rarr from `devel` please? This portion of the code got improved recently.

rcannood:
> @rcannood, before I attempt to reproduce, could you confirm you are using the latest version of Rarr from `devel` please? This portion of the code got improved recently.

Oh I see! That makes sense; let me check with the very latest version of Rarr then.

rcannood:
I was on 2.1.18. I tried 2.1.28, the whole read went from ~920ms / 1.28GB to 466ms / 467MB!

Unfortunately, `read_zarr_array()` still returns a value that's already marked as shared, so `dim(x) <- NULL` duplicates exactly like `as.vector()` did. The total allocation of `read_zarr_sparse_array()` comes out byte-identical on both branches 467,501,224 vs 467,536,056). So `dim(x) <- NULL` won't yet result in a performance boost.

However, I'm happy to merge `dim(x) <- NULL` for when that does get fixed upstream.

Could you finalise this PR by:

- [ ] Rebasing on `devel` (small conflict in `test-Zarr-read.R`)
- [ ] Adding a `NEWS` entry + version bump
- [ ] Adding a comment at the `dim(x) <- NULL` sites mentioning that this is a performance improvement

Bisaloo:
> Merging read_zarr_nullable_integer() / read_zarr_nullable_boolean() into one does make sense, because Rarr already outputs the right vector data types. Could we spin this off into a separate pr?

Let me cherry-pick this into a separate PR and we can put this one on hold for now. I'm not even sure it's something I have control on in Rarr so maybe there is no point in merging.

PR #488, opened 2026-07-21
Update NEWS

rcannood:
The NEWS.md hadn't been updated in a while. This PR adds entries for past PRs.

There are quite a lot of PRs which were merged but aren't in bioc devel yet. After this PR is merged, the devel branch should be pushed to upstream for a sync.


lazappi (approved):
Looks good. The temporary CI fix also needs to be pushed to release to fix that build (just in case you weren't already planning to do that 😸).

rcannood:
Thanks! Indeed; was thinking that it would be good to back port a few fixes. I'll see what i can do

PR #492, opened 2026-07-28
Write and use consolidated metadata

### LINKED ISSUE #461: Consolidated Zarr metadata at the root of the anndata zarr store
### Description

In https://anndata.scverse.org/en/stable/tutorials/zarr-v3.html, it is written:

> All zarr stores are now consolidated by default when written via [anndata.io.write_zarr()](https://anndata.scverse.org/en/stable/generated/anndata.io.write_zarr.html#anndata.io.write_zarr) or [anndata.AnnData.write_zarr()](https://anndata.scverse.org/en/stable/generated/anndata.AnnData.write_zarr.html#anndata.AnnData.write_zarr).

### Proposed Solution

Rarr has now the ability to write consolidated metadata since https://github.com/Huber-group-EMBL/Rarr/pull/184. Would you be interested in a PR implementing this feature?

This should significantly improve performance because a single json file can be read instead of many scattered individual ones. This also gives a relatively easy path towards supporting reading anndata directly from S3.

Bisaloo:
**Related to:** fix #461

## Description

In line of what anndata python is doing, this PR:

- always writes consolidated metadata when writing anndata to Zarr.
- uses the metadata when reading (it's created on the fly if it doesn't exist on disk) to avoid multiple reading of multiple JSON files.

Eventually, in a follow up PR, this can also be used to add reading from S3.

## Checklist

**Before review**

- [ ] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version


Bisaloo:
I'll stop here for a first review & merge if this seems good.

The 2nd stage, which will bring reading from S3 and further performance improvement, will require changing the signature of some helper function, so I prefer to do it in a dedicated PR.

For example, `read_zarr_categorical()` would now have `ordered` in its signature so it can be passed directly from the consolidated metadata, and all `read_zarr_attributes()` calls can be avoided (or even eliminated) from individual low-level helpers.

R/ZarrAnnData.R
```diff
@@ -493,7 +559,18 @@ ZarrAnnData <- R6::R6Class(
...
+      }
+
+      # Create consolidated metadata on the fly, in memory, if it doesn't exist on disk
+      consolidated <- Rarr:::.read_consolidated_metadata(paste0(file, "/"))
+      if (is.null(consolidated)) {
+        raw_consolidated <- Rarr::zarr_consolidate_metadata(file, action = "return")
+        if (private$.zarrformat == 3L) {
+          consolidated <- raw_consolidated$consolidated_metadata
+        }
       }
```
Bisaloo: This is clunky. I'll add some utilities to make it simpler upstream in Rarr. These utilities will be useful for spatialdataR as well.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: anndataR
 Title: AnnData interoperability in R
-Version: 1.3.1
+Version: 1.3.2
```
lazappi, line 3: I'm not sure if we want to bump this yet (I've lost track of how we were managing versions, @rcannood might have a plan)

lazappi (changes requested):
I think this looks good. Does there need to be any checks for the Zarr version?

Bisaloo:
> Does there need to be any checks for the Zarr version?

From Rarr's side, no. Consolidated metadata is generated, and would be beneficial, for both v2 and v3. I don't know if the anndata spec is strict about this but my view is that even if anndata python doesn't generate it / use it for v2, they are not likely to be hindered by it either.

PR #493, opened 2026-07-28
Merge read_zarr_nullable_integer & read_zarr_nullable_boolean

Bisaloo:
**Related to:** #483


lazappi:
I would prefer to explicitly keep separate functions for everything that is defined in the `AnnData` spec. Mostly for consistency but also for potential future versions (i.e. `nullable integer` gets updated to `v0.2.0` but `nullable boolean` doesn't).

Bisaloo:
Ok, I'm going to close this PR if this change is not making a consensus then, thanks for the feedback.

PR #496, opened 2026-07-31
Fix logical matrices losing their shape when written to H5AD

### LINKED ISSUE #495: Writing a logical matrix loses its shape (`hdf5_write_boolean_dataset()`)
`hdf5_write_boolean_dataset()` writes any 2D (or higher) logical matrix as a flat 1D dataset, losing its shape. `read_h5ad()` (R or Python) then fails to reconstruct it.

## Reproduction

```r
m <- matrix(c(TRUE, FALSE, TRUE, FALSE), nrow = 2, ncol = 2)
dim(m)              # 2 2
dim(as.integer(m))  # NULL - as.integer() drops the dim attribute
```

`R/write_hdf5_helpers.R`, `hdf5_write_boolean_dataset()`:

```r
value <- as.integer(value)   # dim(value) becomes NULL here

...
if (!is.null(dim(value))) {  # always FALSE now
  dims <- dim(value)
} else {
  dims <- length(value)      # always taken, matrix flattened to 1D
}
```

Any logical matrix written to `X`/`layers`/`obsm`/`varm` ends up as a 1D dataset instead of a 2D one, and reading it back fails since the shape no longer matches `(n_obs, n_var)`.

## Fix

Capture `dim(value)` before converting to integer, and align `H5Screate_simple()`'s `native` argument with the rest of the write path (`hdf5_write_dataset()` a few lines above uses `native = FALSE`; the boolean writer used `native = TRUE`, which produces a dataset with reversed/wrong dims once the shape is no longer silently discarded). PR incoming with both fixes plus a roundtrip test.

JuanTejedor:
Fixes #495.

`hdf5_write_boolean_dataset()` converted `value` with `as.integer()` before reading `dim(value)` to build the dataspace. `as.integer()` drops the `dim` attribute, so the check always took the flat-vector branch, and any 2D+ logical matrix (`X`/`layers`/`obsm`/`varm`) was written as a 1D dataset — breaking reads since the shape no longer matched `(n_obs, n_var)`.

Fix: capture `dim(value)` before the conversion, and use `native = FALSE` (matching `hdf5_write_dataset()` just above it, and what `write_h5ad_dense_array()`'s existing transpose already assumes).

Added a test writing a logical matrix and checking both the raw on-disk shape and a full roundtrip. Ran the full suite locally (`testthat::test_dir()`): 0 failures from this change (1 pre-existing failure from `Seurat`/`SeuratObject` not being installed in my environment, unrelated).


NEWS.md
```diff
@@ -1,5 +1,6 @@
 # anndataR 1.3.1
 
+- Fix `hdf5_write_boolean_dataset()` writing logical matrices as flat 1D datasets instead of preserving their shape (PR #496).
```
lazappi:
```suggestion
# anndataR devel

- Fix `hdf5_write_boolean_dataset()` writing logical matrices as flat 1D datasets instead of preserving their shape (PR #496).

# anndataR 1.3.1
```

lazappi (changes requested):
This looks good, just a minor thing in the NEWS.

rcannood (approved):
LGTM! Thank you @JuanTejedor!

PR #489, opened 2026-07-22
Backport fixes to Bioc 3.23 - anndataR 1.2.1

rcannood:
Backports fixes from `devel` to `RELEASE_3_23`, limited to changes that fix functionality already present in the 3.23 release (no new features).

- Fix `create_zarr()` so that Zarr stores can be created at relative paths and at paths containing regex metacharacters (PR #478).
- Fix conditional expression in `AbstractAnnData` matrix validation (PR #487).
- Take into account the Rarr 2.1.0 structured datatype breaking change in Zarr tests (PR #462).
- Require Python `anndata<0.13.0` in tests, workflows and vignettes (PR #481).
- Update `CITATION` with the published paper (PR #464).

PR #480, opened 2026-07-15
Add support for `nullable-string-array`

lazappi:
## Description

Add support for the `nullable-string-array` encoding, see https://anndata.scverse.org/en/stable/fileformat-prose.html#nullable-string-specifications-v0-1-0.

Fixes #463 

- [x] Read H5AD
- [x] Write H5AD
- [x] Read Zarr
- [x] Write Zarr

## Checklist

**Before review**

- [ ] Update and regenerate man pages
- [x] Add/update tests
- [ ] Add/update examples in vignettes
- [x] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [ ] Bump devel version

PR #503, opened 2026-08-15
Backport fixes to Bioc 3.23 - anndataR 1.2.2

rcannood:
Backports fixes from `devel` to `RELEASE_3_23`, following the same rule as #489: no new features beyond what fixes compatibility with files Python **anndata** already writes.

- #480
- #496
- #502

Not backported:

- #495
- #387
- #498


lazappi (approved):

dlaehnemann:
@lazappi and @rcannood I was encountering the `nullable-string-array` issue and found my way to this backport of the solution, but realised this is not on Bioconductor as a version `1.2.2`. I don't understand a lot of the infrastructure behind Bioconductor, but I think the issue is probably that the [CI did not run through on the merge-commit](https://github.com/scverse/anndataR/actions/runs/32017240654)?

I would love to have this released on Bioconductor, to have compatibility with `Zarr` and `H5AD` as written by the python `anndata` package.

rcannood:
Oh; you're absolutely right! Let me look into this

PR #520, opened 2026-09-11
Skip tests while rhdf5 2.57.12 warns about NA_character_

rcannood:
**Related to:** Huber-group-EMBL/rhdf5#241, Huber-group-EMBL/rhdf5#242

## Description

rhdf5 2.57.12 warns about `NA_character_` for every numeric dataset containing `NA`/`NaN` that is written into a pre-created dataset (the check lacks a character guard, see the permalink in `known_issues.yaml`). anndataR always pre-creates datasets, so `write_h5ad()` of the test dataset emits 48 of these and the `expect_no_condition()` tests in `test-h5ad-fileclosure.R` fail on every platform.

* add a known-issue entry describing the upstream bug
* skip the three file-closure tests while the entry exists

To revert once rhdf5 with the fix from Huber-group-EMBL/rhdf5#242 is in Bioconductor devel.

## Checklist

**Before review**

- [x] Update and regenerate man pages (n/a)
- [x] Add/update tests
- [x] Add/update examples in vignettes (n/a)
- [ ] Pass CI checks

**Before merge**

- [ ] Update `NEWS`
- [x] Bump devel version (not needed)

Issue #511, opened 2026-08-31
usage_python vignette fails with mudata >= 0.4


rcannood:
The `usage_python` vignette fails on CI since 2026-08-17, breaking R-CMD-check, r-universe and pkgdown on every branch:

```
Quitting from usage_python.Rmd:119-127 [load_mudata_example]
! anndata._io.specs.registry.IORegistryError: No read method registered for IOSpec(encoding_type='AnnData', encoding_version='0.1.0') from <class 'h5py._hl.group.Group'>. You may need to update your installation of anndata.
```

## Cause

mudata 0.4.0 (released 2026-08-17) rewrote `read_h5mu()` to dispatch modality groups through anndata's IO registry on the `encoding-type` attribute. The example file used in the vignette ([minipbcite.h5mu](https://github.com/scverse/anndataR/blob/devel/vignettes/usage_python.Rmd#L119-L127)) was written by mudata <= 0.1.1 and stores `encoding-type: "AnnData"`, while the registry only knows the lowercase `"anndata"` that mudata has written since 0.1.2. mudata's own backwards-compatibility test archives only go back to v0.1.2, so this went unnoticed upstream.

Verified locally:

* mudata 0.3.10 + anndata 0.12: reads the file fine
* mudata 0.4.0 / 0.4.1, with anndata 0.12 **or** 0.13: `IORegistryError`
* mudata 0.4.1 after rewriting the attribute to `"anndata"`: reads the file fine

Minimal reproducible example (no R needed):

```sh
curl -sLO https://github.com/gtca/h5xx-datasets/raw/b1177ac8877c89d8bb355b072164384b4e9cc81d/datasets/minipbcite.h5mu
uv run --with "mudata==0.4.1" python -c "import mudata; mudata.read_h5mu('minipbcite.h5mu')"
```

## Plan

- [ ] Pin `mudata<0.4` in the vignette so CI is green again
- [ ] Report the regression upstream to scverse/mudata -- a fallback for the legacy capitalised `encoding-type` would fix it -- or switch the vignette to an example file written by a recent mudata
- [ ] Remove the pin once either of those lands

PR #510, opened 2026-08-28
Class rename with anndata >= 0.13 and add warning

rcannood:
**Related to:** Fixes #499

## Description

In Python anndata >= 0.13 , `anndata._core.anndata.AnnData` was renamed to `anndata.AnnData`. As a result, the `py_to_r()` S3 method never dispatched and objects coming from Python (#499).

This PR adds support for both variants, but adds a warning when anndata >= 0.13 is detected because there are known issues.

## Checklist

**Before review**

- [x] Update and regenerate man pages
- [x] Add/update tests
- [ ] Add/update examples in vignettes
- [ ] Pass CI checks

**Before merge**

- [x] Update `NEWS`
- [ ] Bump devel version

Issue #499, opened 2026-08-07
AttributeError: 'AnnData' object has no attribute 'as_SingleCellExperiment'


lgatto:
### Description

I can't reproduce the mudata import example from [the *Python Integration with anndataR* vignette](https://anndatar.scverse.org/articles/usage_python.html#multi-modal-data-with-mudata). The error says

```r
Error in py_get_attr(x, name) : 
  AttributeError: 'AnnData' object has no attribute 'as_SingleCellExperiment'
Run `reticulate::py_last_error()` for details.
```

This used to work in March.

### Minimal Reproducible Example

```r
> suppressPackageStartupMessages({
+     library(SingleCellExperiment)
+     library(anndataR)
+     library(reticulate)
+ })
+ > cache <- BiocFileCache::BiocFileCache(ask = FALSE)
+ h5mu_file <- BiocFileCache::bfcrpath(
+   cache,
+   "https://github.com/gtca/h5xx-datasets/raw/b1177ac8877c89d8bb355b072164384b4e9cc81d/datasets/minipbcite.h5mu"
+ )
+ reticulate::py_install("mudata")
+ > Using virtual environment '/home/lgatto/.virtualenvs/r-reticulate' ...
+ /home/lgatto/.virtualenvs/r-reticulate/bin/python -m pip install --upgrade --no-user mudata
Requirement already satisfied: mudata in /home/lgatto/.virtualenvs/r-reticulate/lib/python3.14/site-packages (0.3.10)
Requirement already satisfied: anndata>=0.10.8 in /home/lgatto/.virtualenvs/r-reticulate/lib/python3.14/site-packages (from mudata) (0.13.2)
Requirement already satisfied: h5py in /home/lgatto/.virtualenvs/r-reticulate/lib/python3.14/site-packages (from mudata) (3.16.0)
Requirement already satisfied: numpy in /home/lgatto/.virtualenvs/r-reticulate/lib/python3.14/site-packages (from mudata) (2.5.1)
Requirement already satisfied: pandas>=1.4 in /home/lgatto/.virtualenvs/r-reticulate/lib/python3.14/site-packages (from mudata) (3.0.5)
Requirement already satisfied: scipy in /home/lgatto/.virtualenvs/r-reticulate/lib/python3.14/site-packages (from mudata) (1.18.0)
[...]

PR #523, opened 2026-09-17
Backport mudata pin and anndata 0.13 support to Bioc 3.23 - anndataR 1.2.2

rcannood:
**Related to:** #503, #511, #499

## Description

Backports more fixes from `devel` to `RELEASE_3_23`. 1.2.2 was never pushed to Bioconductor, so these go into 1.2.2 instead of a new version.

The Bioc 3.23 build of 1.2.1 currently fails at `usage_python.Rmd` (`load_mudata_example`) because mudata 0.4 can't read the legacy example `.h5mu` file.

* #512
* #518
* #510
* #462

CI on this branch was already broken before this PR, so it also fixes:

* `test-coverage` hardcoded `bioc-version: devel`, so the release branch was tested against `Rarr` 2.1 -- resolve the version from the branch, like #502 does for `R-CMD-check-bioc`
* the Windows/macOS-ARM setup workaround installed R devel against Bioconductor 3.23 packages -- drop it, as #457 did on `devel`

Not backported, because they only apply to the Bioc devel versions of rhdf5 (>= 2.57) or Rarr (>= 2.1):

* #497
* #519
* #520
* #498

## Checklist

**Before review**

- [x] Add/update tests
- [ ] Pass CI checks

**Before merge**

- [x] Update `NEWS`
- [x] Bump devel version


dlaehnemann:
I would actually like to use this patched version, so I'm curious if anything else needs to happen, for this version `1.2.2` to become available. I would like to package it on `bioconda`, and it seems the regular link usually used for `Bioconductor` packages on `bioconda` is not available yet. Namely, this gives an error:

https://bioconductor.org/packages/3.23/bioc/src/contrib/anndataR_1.2.2.tar.gz

While this link works:

https://bioconductor.org/packages/3.23/bioc/src/contrib/anndataR_1.2.1.tar.gz

Are there some extra CI / build steps that need to happen somewhere? Anything that needs to be triggered manually?

Artur-man:
Take a look at the "Even Odd Schedule" at https://contributions.bioconductor.org/versionnum.html#even-odd-schedule. 

Basically, bioc release versions have fixed package versions so 1.2.1 and 1.2.2 together cannot be available at 3.23 release unfortunately, and I do not believe changes can be backported to the actual release version at git::bioc either, so the `RELEASE_3_23` branch here cannot be pushed to `gitbioc/anndataR@RELEASE_3_23`.

Artur-man:
or not https://contributions.bioconductor.org/git-version-control.html#bug-fix-in-release-and-devel
but this is only the case for the current release branch.

dlaehnemann:
I think the even-odd is for the minor version, so the `Y` in an `X.Y.Z` version specification. So `1.3.2` would be on the `devel` branch, as opposed to `1.2.2` on the release branch. But yeah, I have gotten confused by this versioning before.

Artur-man:
Oh yes, I did not refer to the patch version by even-odd, but rather was trying to say that you cannot access both 1.2.1 and 1.2.2 from release 3.23, does not matter if 1.2.2 is pushed to 3.23 or not.

rcannood:
Hi @dlaehnemann! anndataR 1.2.2 is already on Bioconductor's `RELEASE_3_23` branch. I pushed the branch to Bioc after this merge was done, but the build run hasn't come through yet even though this should be a daily build. I'll wait another for 1.2.2 to land on https://bioconductor.org/checkResults/3.23/bioc-LATEST/anndataR/. If it still isn't there tomorrow, I'll ask bioc devel for an update. At the very least I created a git tag for 1.2.2 already: https://github.com/scverse/anndataR/releases/tag/v1.2.2. This this plan ok with you?



@Artur-man release branches do take bug fixes as long as each push bumps the z version. I'm not entirely sure what your point is.

dlaehnemann:
Yeah, of course. Thanks for the detailed response. I was mainly just lost when trying to figure out where it is hanging, and thus figured this would be the most reasonable place to ask.

In case you are curious where this comes from:

My current workaround (and my current use case) is to actually [install `1.2.1` in a conda environment with all the (optional) dependencies I need](https://github.com/snakemake-workflows/single-cell-counts-preprocessing/blob/213c6e794be869f0cd2ae67ff3cd1c8dffec56df/workflow/envs/soupx.yaml#L7), and [then install a particular commit from GitHub into that conda environment with `devtools`](https://github.com/snakemake-workflows/single-cell-counts-preprocessing/blob/213c6e794be869f0cd2ae67ff3cd1c8dffec56df/workflow/envs/soupx.post-deploy.sh#L6). So this works for me, but simply pointing to `1.2..2` would be even cleaner.

Artur-man:
> I'm not entirely sure what your point is. 

@dlaehnemann was trying to download 1.2.1 and 1.2.2 (or multiple versions) from release 3.23, was explaining that cannot happen, but I was incorrect about bug fixes to 3.23 of course, my bad.

rcannood:
@dlaehnemann I see, that's a good solution! 

anndataR 1.2.2 is now live: https://www.bioconductor.org/packages/release/bioc/html/anndataR.html

I see https://github.com/bioconda/bioconda-recipes/pull/69464 was created -- looks good to me!

